PR #83, opened 2020-11-23
remove dynamic memory allocation in ota cbor

pvyawaha:
*Issue #, if available:*

*Description of changes:*
This change removes the dynamic allocation in ota cbor component.


source/ota_cbor.c
```diff
@@ -48,7 +69,8 @@
  * @param[out] pBlockId Decoded block id value.
  * @param[out] pBlockSize Decoded block size value.
  * @param[out] pPayload Buffer for the decoded payload.
```
tgsong, line 71: so `pPayload` is now preallocated and passed in by user? We should then change its type to `uint8_t *` instead of `uint8_t **`.

tgsong (changes requested):

source/ota_cbor.c
divekarshubham: This will require a change in the otaDataInterface (.decodeFileBlock) touching multiple files. We could have a separate PR for this.

divekarshubham (dismissed):

divekarshubham (approved):

cobusve (approved):

tgsong (approved):

PR #85, opened 2020-11-24
Renaming OTA API

divekarshubham:
*Issue #, if available:*

*Description of changes:*
* Renamed `OTA_AgentInit()` to `OTA_Init()` 
* Added `OTA_GetStatistics()` to replace `OTA_GetPacketsReceived` `OTA_GetPacketsQueued` `OTA_GetPacketsProcessed` `OTA_GetPacketsDropped`
* Renamed `OTA_GetAgentState()` to `OTA_GetState()` 
* Renamed `OTA_AgentShutdown()` to `OTA_Shutdown()`

* Modifications to the Unit Tests to include new API changes


tgsong (dismissed):

source/include/ota.h
```diff
@@ -688,17 +688,17 @@ void otaAgentTask( const void * pUnused );
...
  * @brief Get the number of OTA message packets received by the OTA agent.
  *
- * @note Calling @ref OTA_AgentInit will reset this statistic.
+ * @note Calling @ref OTA_Init will reset this statistic.
  *
  * @return The number of OTA packets that have been received but not
  * necessarily queued for processing by the OTA agent.
  */
-uint32_t OTA_GetPacketsReceived( void );
+uint32_t OTA_GetStatistics( void );
```
pvyawaha, line 696: The OTA_GetStatistics is supposed to return all four params and not just packest received.

divekarshubham: Done,  change in https://github.com/divekarshubham/ota-for-aws-iot-embedded-sdk-1/commit/c3d146e6feacdaab0777e2a4a6391205088784d5

tgsong (approved):

dan4thewin (dismissed):

source/ota.c
```diff
@@ -190,6 +190,31 @@ static bool otaClose( OtaFileContext_t * const pFileContext );
 static OtaErr_t setImageStateWithReason( OtaImageState_t stateToSet,
                                          uint32_t reasonToSet );
 
+/* The default OTA callback handler if not provided to OTA_Init(). */
```
tgsong, line 193: These files are not used anymore and already deleted.

tgsong (changes requested):

source/ota.c
divekarshubham: Thanks! Removed in https://github.com/aws/ota-for-aws-iot-embedded-sdk/pull/85/commits/6a8d1b7eee8715f15deecd3fc26ad8f23662a3c7

tgsong (approved):

source/ota.c
```diff
@@ -2772,41 +2771,18 @@ OtaState_t OTA_AgentShutdown( uint32_t ticksToWait )
...
-    return otaAgent.statistics.otaPacketsProcessed;
-}
-
-/*
- * Return the number of packets received.
- */
-uint32_t OTA_GetPacketsReceived( void )
-{
-    return otaAgent.statistics.otaPacketsReceived;
+    *pStatistics = otaAgent.statistics;
```
gkwicker, line 2784: Since this is a public API, you need to validate `pStatistics` before dereferencing it, e.g.:

```
OtaErr_t returnStatus = OTA_ERR_INVALID_PARAMS;
if( pStatistics != NULL )
{
    *pStatistics = otaAgent.statistics;
    returnStatus = OTA_ERR_NONE;
}
return returnStatus;

File: test/unit-test/ota_utest.c [hunk 4]
```diff
@@ -370,7 +370,7 @@ static void otaInitDefault()
 static void otaDeinit()
 {
     mockOSEventReset( NULL );
-    OTA_AgentShutdown( 1 );
+    OTA_Shutdown( 1 );
```
gkwicker, line 373: What does `1` mean?  If it means "true" then let's pass `true` here.  And a deeper question – what does it mean to pass a value other than 1 to this API?  If different values have different meanings, should that parameter be an `enum` type or should we use `#define`s for the values?

test/unit-test/ota_utest.c
```diff
@@ -657,82 +657,85 @@ void test_OTA_ResumeWhenStopped()
...
     otaGoToState( OtaAgentStateReady );
-    TEST_ASSERT_EQUAL( OtaAgentStateReady, OTA_GetAgentState() );
+    TEST_ASSERT_EQUAL( OtaAgentStateReady, OTA_GetState() );
 
-    TEST_ASSERT_EQUAL( 0, OTA_GetPacketsDropped() );
-    TEST_ASSERT_EQUAL( 0, OTA_GetPacketsQueued() );
-    TEST_ASSERT_EQUAL( 0, OTA_GetPacketsProcessed() );
-    TEST_ASSERT_EQUAL( 0, OTA_GetPacketsReceived() );
+    OtaAgentStatistics_t statistics = { 0 };
+    TEST_ASSERT_EQUAL( OTA_ERR_NONE, OTA_GetStatistics( &statistics ) );
```
gkwicker, line 704: We need to test both success and failure case (see comment above) here.  Also, lines 704-709 aren't verifying anything about what `OTA_GetStatistics()` has done since `statistics` was initialized to 0.  `statistics` should be initialized to something other than the values returned by `OTA_GetStatistics()`, and this test should initialize each member in the returned `OtaAgentStatistics_t` object such that lines 704-709 can verify that each was assigned correctly.

gkwicker (changes requested):
See comments

test/unit-test/ota_utest.c
divekarshubham: `OTA_Shutdown()` takes the parameter `ticksToWait`, which is the time to spend idle before the agent shuts down.
For the calls to the shutdown we have a const parameter `otaDefaultWait` [1000]
Updated this line to use the default wait.

source/ota.c
divekarshubham: Thanks, updated:
https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/9b8ba39796aab5af8d52b3423c4a497ae0fbbe4a/source/ota.c#L2782

test/unit-test/ota_utest.c
divekarshubham: Added check for failure: https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/9b8ba39796aab5af8d52b3423c4a497ae0fbbe4a/test/unit-test/ota_utest.c#L703

We would add further tests to check the correctness of the API with the OTA Tests PR.

gkwicker: This will be fixed in a separate PR

gkwicker (approved):

pvyawaha (approved):

PR #88, opened 2020-11-25
Use pal interface and fix memory leak issues

pvyawaha:
*Issue #, if available:*

*Description of changes:*
This change fixes the pal interface calls and memory leaks along with refactoring for dependency.

### DIFF
--- a/source/include/ota_private.h
+++ b/source/include/ota_private.h
@@ -144,9 +144,10 @@ typedef enum
     IngestResultBadData = -8,            /*!< The data block from the server was malformed. */
     IngestResultWriteBlockFailed = -9,   /*!< The PAL layer failed to write the file block. */
     IngestResultNullResultPointer = -10, /*!< The pointer to the close result pointer was null. */
+    IngestResultNoDecodeMemory = -11,    /*!< Memory could not be allocated for decoding . */
     IngestResultUninitialized = -127,    /*!< Software BUG: We forgot to set the result code. */
     IngestResultAccepted_Continue = 0,   /*!< The block was accepted and we're expecting more. */
-    IngestResultDuplicate_Continue = 1,  /*!< The block was a duplicate but that's OK. Continue. */
+    IngestResultDuplicate_Continue = 1   /*!< The block was a duplicate but that's OK. Continue. */
 } IngestResult_t;
 
 /**
--- a/source/include/ota_private.h
+++ b/source/include/ota_private.h
@@ -49,6 +49,22 @@
  * in ota_config.h file. */
 #include "ota_config_defaults.h"
 
+/**
+ * @cond DOXYGEN_IGNORE
+ * Doxygen should ignore this section.
+ */
+
+/* bool is defined in only C99+. */
+#if defined( __cplusplus ) || ( defined( __STDC_VERSION__ ) && ( __STDC_VERSION__ >= 199901L ) ) || \
+    ( defined( _MSC_VER ) && ( _MSC_VER >= 1800 ) )
+    #include <stdbool.h>
+#elif !defined( bool )
+    #define bool     int8_t
+    #define false    ( int8_t ) 0
+    #define true     ( int8_t ) 1
+#endif
+/** @endcond */
+
 /* General constants. */
 #define LOG2_BITS_PER_BYTE           3U                                                   /*!< Log base 2 of bits per byte. */
 #define BITS_PER_BYTE                ( ( uint32_t ) 1U << LOG2_BITS_PER_BYTE )            /*!< Number of bits in a byte. This is used by the block bitmap implementation. */
--- a/source/include/ota_private.h
+++ b/source/include/ota_private.h
@@ -49,12 +49,6 @@
  * in ota_config.h file. */
 #include "ota_config_defaults.h"
 
-#include "ota.h"
-#include "ota_os_interface.h"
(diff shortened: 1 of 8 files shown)


tgsong (dismissed):

tgsong (dismissed):

source/ota.c
```diff
@@ -1265,6 +1013,78 @@ static OtaErr_t jobNotificationHandler( const OtaEventData_t * pEventData )
     return ( OTA_SignalEvent( &eventMsg ) == true ) ? OTA_ERR_NONE : OTA_ERR_EVENT_Q_SEND_FAILED;
 }
 
+static void freeFileContextMem( OtaFileContext_t * const pFileContext )
+{
+    if( pFileContext != NULL )
+    {
+        /* Free or clear the filepath buffer.*/
+        if( ( pFileContext->pFilePath != NULL ) && ( pFileContext->filePathMaxSize > 0 ) )
```
dan4thewin, line 1021: It appears in all of these conditionals that a somethingMaxSize of 0 is supposed to indicate heap was used.  Still, wouldn't free be called here is the pointer == NULL ?

source/ota.c
```diff
@@ -1772,62 +1580,62 @@ static OtaJobParseErr_t parseJobDocFromCustomCallback( const char * pJson,
...
+            /* Custom job was parsed by external callback successfully. Grab the job name from the file
+             *  context and save that in the ota agent */
+            if( strlen( pFileContext->pJobName ) > 0 )
             {
-                LogError( ( "Failed to update job status: "
-                            "updateJobStatus returned error: "
-                            "OtaErr_t=%d",
-                            otaErr ) );
-            }
+                memcpy( otaAgent.pActiveJobName, otaAgent.fileContext.pJobName, OTA_JOB_ID_MAX_SIZE );
```
dan4thewin, line 1596: If you're going to count the string length, shouldn't you use that here instead of OTA_JOB_ID_MAX_SIZE?

source/ota.c
```diff
@@ -1772,62 +1580,62 @@ static OtaJobParseErr_t parseJobDocFromCustomCallback( const char * pJson,
...
-            LogInfo( ( "Job document parsed from external callback" ) );
+                /* Everything looks OK. Set final context structure to start OTA. */
+                **pFinalFile = *pFileContext;
+                LogInfo( ( "Job document parsed from external callback" ) );
 
-            /* We don't need the job name memory anymore since we're done with this job. */
-            free( otaAgent.pOtaSingletonActiveJobName );
-            otaAgent.pOtaSingletonActiveJobName = NULL;
+                /* We don't need the job name memory anymore since we're done with this job. */
+                memset( otaAgent.pActiveJobName, 0, OTA_JOB_ID_MAX_SIZE );
```
dan4thewin, line 1607: I see `otaAgent.pActiveJobName` used in a lot of functions.  How did we go from populating it 10 lines earlier to being done with it here?

source/include/ota_private.h
```diff
@@ -82,18 +92,31 @@
...
  *
  */
-#define OTA_NUM_JOB_PARAMS              ( 21 )
+#define OTA_NUM_JOB_PARAMS             ( 21 )
+
+/**
+ * @brief Maximum size of the Job ID.
+ *
+ */
+#define OTA_JOB_ID_MAX_SIZE            ( 72UL + 1UL )
```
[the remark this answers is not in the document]
pvyawaha: This is AFR job so AFR_OTA- is prefix attached to it

source/ota.c
```diff
@@ -478,7 +439,7 @@ static OtaErr_t setImageStateWithReason( OtaImageState_t stateToSet,
     /* Now update the image state and job status on service side. */
     otaAgent.imageState = state;
 
-    if( otaAgent.pOtaSingletonActiveJobName != NULL )
+    if( strlen( otaAgent.pActiveJobName ) > 0 )
```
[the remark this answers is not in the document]
pvyawaha: This can be user initiated setimage state by calling function OTA_SetImageState , so it can be called when no active job is available, so the expected behavior is only platform state is updated.

source/ota.c
pvyawaha: Here the job status is updated with success( based on the custom job parsing ) and thats why we dont need the context anymore.

source/ota.c
pvyawaha: Updated all three occurrences 7d00faf

source/ota.c
pvyawaha: Std free called with NULL should not harm , but I have updated the if else conditions in 7d00faf

source/include/ota_private.h
```diff
@@ -144,9 +167,10 @@ typedef enum
     IngestResultBadData = -8,            /*!< The data block from the server was malformed. */
     IngestResultWriteBlockFailed = -9,   /*!< The PAL layer failed to write the file block. */
     IngestResultNullResultPointer = -10, /*!< The pointer to the close result pointer was null. */
+    IngestResultNoDecodeMemory = -11,    /*!< Memory could not be allocated for decoding . */
```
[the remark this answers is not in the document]
pvyawaha: added comment in 7d00faf

tgsong (approved):

dan4thewin (approved):

PR #89, opened 2020-11-26
Bug fixes in #88

tgsong:
*Description of changes:*

- Fix include of stdbool.h
- decodeFileBlock should accept const network buffer
- decodeFileBlock_mqtt should not reuse network buffer
- remove unused methods
- signature buffer should not be dynamically allocated
- only update http url when it's not NULL
- decoded data payload should be freed after processDataBlock

### DIFF
--- a/source/ota.c
+++ b/source/ota.c
@@ -29,6 +29,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <errno.h>
+#include <assert.h>
 
 /* OTA agent includes. */
 #include "ota.h"
@@ -95,7 +96,7 @@ static OtaDataInterface_t otaDataInterface;
 /* Called when the OTA agent receives a file data block message. */
 
 static IngestResult_t ingestDataBlock( OtaFileContext_t * pFileContext,
-                                       uint8_t * pRawMsg,
+                                       const uint8_t * pRawMsg,
                                        uint32_t messageSize,
                                        OtaErr_t * pCloseResult );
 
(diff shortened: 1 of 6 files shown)


source/ota.c
```diff
@@ -1262,45 +1155,34 @@ static DocParseErr_t decodeAndStoreKey( char * pValueInJson,
...
+                    "base64Decode returned error: "
+                    "error=%d",
+                    base64Status ) );
+        err = DocParseErrBase64Decode;
     }
     else
     {
-        /* We failed to allocate needed memory. Everything will be freed below upon failure. */
-        err = DocParseErrOutOfMemory;
+        ( *pSig256 )->size = ( uint16_t ) actualLen;
```
yanjos-dev, line 1182: Is this case not considered an error? Also, if actualLen is always supposed to be zero here, then we may just want to assign this to zero directly. This way it's more explicit that this is the behavior during an error.

tgsong: What do you mean? This is the success path right? `base64Status == Base64Success` in this else branch.

source/ota.c
```diff
@@ -1789,15 +1671,20 @@ static OtaJobParseErr_t verifyActiveJobStatus( OtaFileContext_t * pFileContext,
...
+                {
+                    /* The buffer is allocated by us, free first then update. */
+                    otaAgent.pOtaInterface->os.mem.free( otaAgent.fileContext.pUpdateUrlPath );
+                    otaAgent.fileContext.pUpdateUrlPath = pFileContext->pUpdateUrlPath;
+                    pFileContext->pUpdateUrlPath = NULL;
+                }
+                else
+                {
+                    /* The buffer is provided by user, directly copy the new url to it. */
+                    memcpy( otaAgent.fileContext.pUpdateUrlPath, pFileContext->pUpdateUrlPath, otaAgent.fileContext.updateUrlMaxSize );
```
yanjos-dev, line 1686: Somewhat unrelated to this PR, so it could be considered in a separate PR, but shouldn't we be checking that the amount of data we are trying to copy is less than the max amount we can store? I see that it won't be possible to write outside the bounds of the array, but we may not be copying the complete URL.

pvyawaha (approved):

source/ota.c
```diff
@@ -1262,45 +1155,34 @@ static DocParseErr_t decodeAndStoreKey( char * pValueInJson,
...
 {
     DocParseErr_t err = DocParseErrNone;
+    size_t actualLen = 0;
+    Base64Status_t base64Status = 0;
+    Sig256_t ** pSig256 = pParamAdd;
 
-    /* Allocate space for and decode the base64 signature. */
-    void * pSignature = otaAgent.pOtaInterface->os.mem.malloc( sizeof( Sig256_t ) );
+    /* pSig256 should point to pSignature in OtaFileContext_t, which is statically allocated. */
+    assert( *pSig256 != NULL );
```
yanjos-dev, line 1163: Is there an upper bound for the size of a signature that we can receive in a job document? I didn't think that there was.

source/ota.c
tgsong: destination is `otaAgent.fileContext.pUpdateUrlPath`, and size we're copying is `otaAgent.fileContext.updateUrlMaxSize`, which is the buffer size of that destination pointer. `pFileContext->pUpdateUrlPath` I believe is pointing to network buffer.

source/ota.c
tgsong: See `kOTA_MaxSignatureSize`, it's part of `Sig256_t` struct.

source/ota.c
yanjos-dev: You're right, I misread the condition

PR #90, opened 2020-11-26
MISRA warnings

tgsong:
*Issue #, if available:*

*Description of changes:*
Fix MISRA warnings in OTA.


source/include/ota_os_interface.h
```diff
@@ -213,7 +213,9 @@ typedef struct OtaTimerInterface
  */
 typedef struct OtaMallocInterface
 {
+    /* coverity[misra_c_2012_rule_21_3_violation] This is part of our interface . */
```
dan4thewin, line 216: Still too terse, but given the blanket nature of this deviation I recommend you put this in your misra.config instead.

test/unit-test/ota_os_posix_utest.c
```diff
@@ -32,48 +32,36 @@
...
-        retVal = timerAttr.it_value.tv_nsec;
-    }
+    timerCallbackInovked = true;
+    clock_gettime( CLOCK_REALTIME, &now );
+    int elapse_ms = ( now.tv_sec - start.tv_sec ) * 1e3 + ( now.tv_nsec - start.tv_nsec ) / 1e6;
 
-    return retVal;
+    /* Assert the time elapsed is within 0.9 to 1.1 of the specified time. */
+    TEST_ASSERT_LESS_THAN( OTA_DEFAULT_TIMEOUT * 1.1, elapse_ms );
+    TEST_ASSERT_GREATER_THAN( OTA_DEFAULT_TIMEOUT * 0.9, elapse_ms );
```
dan4thewin, line 64: What is this testing?  It looks brittle.

source/ota_mqtt.c
```diff
@@ -292,21 +300,17 @@ static OtaErr_t unsubscribeFromDataStream( const OtaAgentContext_t * pAgentCtx )
...
-    const OtaFileContext_t * pFileContext = 0;
+    const OtaFileContext_t * pFileContext = NULL;
 
     assert( pAgentCtx != NULL );
 
     pFileContext = &( pAgentCtx->fileContext );
 
     /* Try to build the dynamic data stream topic and unsubscribe from it. */
-    topicLen = ( uint16_t ) snprintf( pOtaRxStreamTopic, /*lint -e586 Intentionally using snprintf. */
+    /* coverity[misra_c_2012_rule_21_6_violation] Buffer overlap will never happen. */
```
dan4thewin, line 312: Why use snprintf()?  Seems like you could use strncpy() instead.

dan4thewin (changes requested):

test/unit-test/ota_os_posix_utest.c
tgsong: It's testing the timer implementation, one of the test starts the timer with timeout of 1000 ms, this is asserting the time elapsed is between 900 ms and 1100 ms. Before  the test requires direct access to the static variable defined in os posix, so I changed to this way. If you have other suggestion I would like to hear.

source/ota_mqtt.c
tgsong: Because we're not just handling strings, see here, https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/b76149ae0737a04cc48fa99b79a9257211e0e8c1/source/ota_mqtt.c#L63-L75. Using snprintf would handle all cases, and we know the undefined behavior can never happen here.

source/include/ota.h
```diff
@@ -58,6 +58,7 @@
  * @brief The OTA signature algorithm string is specified by the PAL.
  *
  */
+/* coverity[misra_c_2012_rule_8_6_violation] This is defined in OTA pal. */
```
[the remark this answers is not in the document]
tgsong: Updated comments in eb1c445.

source/include/ota_os_interface.h
tgsong: Updated comments in eb1c445.

source/portable/os/ota_os_posix.c
```diff
@@ -27,7 +27,10 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
-/* coverity[misra_c_2012_rule_21_5_violation] This is on linux. */
+
+/* MISRA rule 21.5 prohibits use of signal.h on embedded system. But this implementation is on linux.
+ * So it's a false positive. */
```
dan4thewin, line 32: I don't see anything that says this rule is only for embedded systems.  Can you provide a reference?

source/ota_mqtt.c
```diff
@@ -226,6 +226,12 @@ static OtaErr_t subscribeToJobNotificationTopics( const OtaAgentContext_t * pAge
     assert( pAgentCtx != NULL );
 
     /* Build and subscribe to the first topic. */
+
+    /* MISRA rule 21.6 prohibits use of all functions from stdio.h because they have undefined
+     * behavior. So snprintf, undefined behavior only occurs when the destination buffer and format
+     * buffer overlaps. We are using static buffers and format strings defined in this files, and
```
dan4thewin, line 232: Can you provide a reference for the assertion that undefined behavior is limited to overlaps and only overlaps?

source/portable/os/ota_os_posix.c
```diff
@@ -402,13 +409,19 @@ OtaErr_t Posix_OtaDeleteTimer( OtaTimerId_t otaTimerId )
 void * STDC_Malloc( size_t size )
 {
     /* Use standard C malloc.*/
-    /* coverity[misra_c_2012_rule_21_3_violation] This is on linux. */
+
+    /* MISRA rule 21.3 prohibits the use of malloc and free from stdlib.h. But this implementation
+     * is on linux, so it's a false positive. */
```
dan4thewin, line 414: I understand that use of malloc is ubiquitous on systems with virtual memory, but I don't see how that makes it a false positive.  There are three listed examples of undefined behavior in the rationale section of rule 21.3.  I think it would be better to justify the deviation by explaining why these behaviors do not pertain.

dan4thewin (changes requested):

test/unit-test/ota_os_posix_utest.c
dan4thewin: On the face of it, it looks reasonable, and yet this test will run on systems that are probably not real-time, and may indeed be virtual.  Also, this provides only a single sample of the latency of a single timer.  If timer latency is a real concern, then this test is not thorough enough.  If it's not a real concern, then why test it?

source/portable/os/ota_os_posix.c
tgsong: I believe MISRA itself is for embedded system, quote from themselves,

MISRA, originally created as the Motor Industry Software Reliability Association, is a consortium concerned with best practice in developing safety- and security-related embedded systems.

source/ota_mqtt.c
tgsong: Quote from C standard 7.19.6.5,

The snprintf function is equivalent to fprintf, except that the output is written into
an array (specified by argument s) rather than to a stream. If n is zero, nothing is written,
and s may be a null pointer. Otherwise, output characters beyond the n-1st are
discarded rather than being written to the array, and a null character is written at the end
of the characters actually written into the array. **If copying takes place between objects
that overlap, the behavior is undefined**.

test/unit-test/ota_os_posix_utest.c
tgsong: This test only runs on linux, it's for posix implementation only.

source/portable/os/ota_os_posix.c
tgsong: Because this implementation only run on linux desktop, it's not for embedded devices running RTOS. I don't see why we're not allowed to use malloc here, otherwise how would you recommend implementing the malloc interface on linux?

yanjos-dev (approved):

divekarshubham (dismissed):

source/ota_mqtt.c
dan4thewin: Have a look at this.  It includes links to Annex J.2 of the standard where undefined behaviors are listed.
https://www.misra.org.uk/forum/viewtopic.php?t=1725

source/portable/os/ota_os_posix.c
dan4thewin: What about embedded linux systems?

source/portable/os/ota_os_posix.c
dan4thewin: I'd say, don't use dynamic allocation at all.  Barring that, I'd say use malloc, but explain how you use is safe (CBMC proofs?).  Also, wouldn't we want embedded linux users to take advantage of this code?

test/unit-test/ota_os_posix_utest.c
dan4thewin: I agree, but that doesn't answer my question.

source/portable/os/ota_os_posix.c
pvyawaha: The design decision for OTA library is to use user allocated buffers but if the buffer is NULL/Zero then dynamic allocation si done.

pvyawaha: The design decision for OTA library is to use buffers passed by ota demo/application and when the buffer is NULL/zero then only use dynamic allocation. As this is in the POSIX OS port functions which are passed to the OTA library and user can override it to not use dynamic allocation at all.

source/portable/os/ota_os_posix.c
cobusve: MISRA is really a set of rules to improve the quality of code written in C, it does not only apply to embedded code, it applies to all C-code. Technically standards like ISO 26262 and IEC 61508 require C code written as part of any critical system use a "safe subseet" of the language. This requirement is met by following the MISRA rules. If the system is written in C and runs on Linux this would equally apply.

source/ota_mqtt.c
cobusve: What is the defined behavior for snprintf when the number of parameters do not match the tokens in the format specifier, also what if the types do not match?

Some references (this is a small subset of the full list, really why these functions are outlawed):
C99 section 7.19.6.1 - 
Par 2 "If there are insufficient arguments for the format, the behavior is undefined."
Par 4 "If a precision appears with any other conversion specifier, the behavior is undefined."
Par 6 "For g and G conversions, trailing zeros are not removed from the result. For other conversions, the behavior is undefined."
Par 6 "... if a precision is specified, the 0 flag is ignored. For other conversions, the behavior is undefined."
Par 7 "...If a length modifier appears with any conversion specifier other than as specified above, the behavior is undefined."
Par 9 "If a conversion specification is invalid, the behavior is undefined."
Par 9 "If any argument is
not the correct type for the corresponding conversion specification, the behavior is undefined."

divekarshubham (dismissed):

yanjos-dev (approved):

dan4thewin (approved):
I'd like there to be more follow-up discussion, especially about snprintf(); but, since I deem this a two-way door, I'm approving it.

PR #93, opened 2020-12-01
HTTP unit tests

tgsong:
**Note: compiler warnings check is failing because of pending change on error enums, which will be a separate PR.**

*Description of changes:*
- Turn on HTTP in unit tests config and add OTA HTTP unit tests
- Fix compiler warnings in OTA http
- Fix a bug in OTA http decode
- Add log level control for unit tests (so that compiler doesn't warn about variables only used in logs)
- Remove cmock from unit tests cmake file
- Add compiler warnings to CI checks
- Separate unit tests and coverage check (because in unit tests we want to turn on assert while in coverage we want it off)
- Clean up coverage data from cmake instead of in GitHub CI file


divekarshubham (dismissed):

test/unit-test/ota_config.h
```diff
@@ -31,6 +31,28 @@
 #ifndef OTA_CONFIG_H_
 #define OTA_CONFIG_H_
 
-/* Dummy ota_config.h for building the library. */
+#define configENABLED_DATA_PROTOCOLS    ( OTA_DATA_OVER_MQTT | OTA_DATA_OVER_HTTP )
```
yanjos-dev, line 34: Why are we defining print functions for a unit test?

test/unit-test/ota_utest.c
```diff
@@ -51,19 +51,21 @@
...
-#define JOB_DOC_SELF_TEST_LENGTH     ( sizeof( JOB_DOC_SELF_TEST ) - 1 )
+#define JOB_DOC_HTTP                 "{\"clientToken\":\"0:testclient\",\"timestamp\":1602795143,\"execution\":{\"jobId\":\"AFR_OTA-testjob22\",\"status\":\"QUEUED\",\"queuedAt\":1602795128,\"lastUpdatedAt\":1602795128,\"versionNumber\":1,\"executionNumber\":1,\"jobDocument\":{\"afr_ota\":{\"protocols\":[\"HTTP\"],\"files\":[{\"filepath\":\"/test/demo\",\"filesize\":" OTA_TEST_FILE_SIZE_STR ",\"fileid\":0,\"certfile\":\"test.crt\",\"update_data_url\":\"https://dummy-url.com/ota.bin\",\"auth_scheme\":\"aws.s3.presigned\",\"sig-sha256-ecdsa\":\"MEQCIF2QDvww1G/kpRGZ8FYvQrok1bSZvXjXefRk7sqNcyPTAiB4dvGt8fozIY5NC0vUDJ2MY42ZERYEcrbwA4n6q7vrBg==\"}] }}}}"
 #define JOB_DOC_INVALID              "not a json"
-#define JOB_DOC_INVALID_LENGTH       ( sizeof( JOB_DOC_INVALID ) - 1 )
 
 /* OTA application buffer size. */
 #define OTA_UPDATE_FILE_PATH_SIZE    100
```
yanjos-dev, line 60: Where are these numbers coming from? It's not clear if these are max sizes or sizes of example strings.

test/unit-test/ota_utest.c
```diff
@@ -356,6 +381,8 @@ static void otaInit( const char * pClientID,
     pOtaAppBuffer.certFilePathSize = OTA_CERT_FILE_PATH_SIZE;
     pOtaAppBuffer.pStreamName = pOtaAppBuffer.pCertFilePath + pOtaAppBuffer.certFilePathSize;
     pOtaAppBuffer.streamNameSize = OTA_STREAM_NAME_SIZE;
+    pOtaAppBuffer.pUrl = pOtaAppBuffer.pStreamName + pOtaAppBuffer.streamNameSize;
```
yanjos-dev, line 384: Is this corrrect? It seems odd that we are assigning a string to a stream name plus the size of the stream.

test/unit-test/ota_utest.c
```diff
@@ -434,6 +475,12 @@ static void otaGoToStateWithTimeout( OtaState_t state,
         otaInitDefault();
     }
 
+    /* Default to the MQTT job doc. */
+    if( pOtaJobDoc == NULL )
+    {
+        pOtaJobDoc = JOB_DOC_A;
```
yanjos-dev, line 481: does this need to be a global?

test/unit-test/ota_utest.c
```diff
@@ -902,7 +959,26 @@ void test_OTA_ReceiveFileBlockEmpty()
...
+    pOtaJobDoc = JOB_DOC_HTTP;
+
+    otaGoToState( OtaAgentStateWaitingForFileBlock );
+    TEST_ASSERT_EQUAL( OtaAgentStateWaitingForFileBlock, OTA_GetState() );
+
+    otaInterfaces.os.event.send = mockOSEventSend;
+
+    otaEvent.eventId = OtaAgentEventReceivedFileBlock;
+    otaEvent.pEventData = &eventBuffer;
+    otaEvent.pEventData->dataLength = 2 * OTA_FILE_BLOCK_SIZE;
```
yanjos-dev, line 975: It may be better to set this to the block size + 1. Setting it to 2x the expected block size doesn't check for off by one bugs in the code.

test/unit-test/ota_utest.c
```diff
@@ -966,10 +1042,56 @@ void test_OTA_ReceiveFileBlockComplete()
...
+     * because we're receiving multiple blocks in this test. */
+    otaInterfaces.os.event.send = mockOSEventSend;
+
+    /* Fill the file block. */
+    for( idx = 0; idx < sizeof( pFileBlock ); idx++ )
+    {
+        pFileBlock[ idx ] = idx % UINT8_MAX;
+    }
+
+    while( remainingBlocks > OTA_FILE_BLOCK_SIZE )
```
yanjos-dev, line 1066: Can you rename the remainingBlocks variable? This is a confusing name. It is not the number of blocks that are remaining, it is the number of bytes remaining (assuming that OTA_FILE_BLOCK_SIZE is in terms of bytes). Alternatively this could be changed to actually represent the number of remaining blocks by initializing it to the file size mod the block size, then changing how it's decremented.

test/unit-test/ota_utest.c
```diff
@@ -966,10 +1042,56 @@ void test_OTA_ReceiveFileBlockComplete()
...
+    {
+        otaEvent.pEventData = &eventBuffer;
+        memcpy( otaEvent.pEventData->data, pFileBlock, OTA_FILE_BLOCK_SIZE );
+        otaEvent.pEventData->dataLength = OTA_FILE_BLOCK_SIZE;
+
+        OTA_SignalEvent( &otaEvent );
+        otaWaitForEmptyEvent();
+        TEST_ASSERT_EQUAL( OtaAgentStateWaitingForFileBlock, OTA_GetState() );
+
+        /* TODO, statistics is now broken. Need to fix it to test OTA_GetPacketsReceived
```
yanjos-dev, line 1076: Can we fix this in this PR? I don't think we should have todo statements in code that is being merged into the main branch.

yanjos-dev (changes requested):

test/unit-test/ota_config.h
tgsong: It's for compiler warnings, without it we will have some unused variable warnings.

test/unit-test/CMakeLists.txt
```diff
@@ -51,20 +34,12 @@ list(APPEND test_include_directories ".")
...
-    "${MODULE_ROOT_DIR}/tools/cmock/project.yml"
-    "${mock_include_list}"
-    "${mock_define_list}"
-)
-
 create_real_library(${real_name}
     "${real_source_files}"
     "${real_include_directories}"
-    "${mock_name}"
+    ""
```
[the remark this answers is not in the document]
tgsong: that's a mistake, will deleted

test/unit-test/ota_utest.c
tgsong: it's just some arbitrary numbers I made up. Since we don't restrict this from API, user can pass arbitrary size of buffer too. But I do see some action items to add minimum size required in our API, which should be a separate PR.

test/unit-test/ota_utest.c
tgsong: I'm just making pointers to point to different offset in the buffer I declared above.

test/unit-test/ota_utest.c
tgsong: `pOtaJobDoc`? It can be a parameter to `otaReceiveJobDocument`, but I did this way to control if I'm testing MQTT or HTTP, also avoid changing the signature of `otaGoToState`.

test/unit-test/ota_utest.c
tgsong: This is still broken, I just tried, all statistics are 0. Since it's in `ota.c`, I think it's better to fix in another PR.

test/unit-test/ota_utest.c
tgsong: will use `remainingBytes` instead.

test/unit-test/CMakeLists.txt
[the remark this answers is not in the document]
tgsong: actually no, it's required by the `create_real_library` function

divekarshubham (approved):

yanjos-dev (approved):

cobusve (approved):

PR #94, opened 2020-12-01
Replace error macros with enums

tgsong:
*Description of changes:*
The goal is to get rid of the circular includes between `ota.h` and interface headers. Basically `OtaErr_t` is defined in ota.h and required these interface headers,

- `ota_mqtt_interface.h`
- `ota_http_interface.h`
- `ota_os_interface.h`
- `ota_platform_interface`

The proposal in this PR is to have separate error enum defined for each interface instead of using the `OtaErr_t` from ota.h. Because we still want the underlying integers in these error enums to have different values so that we can compose them together to report to the cloud, we define ranges for each error enums,

- OTA agent: 0 - 0x7F
- OTA OS interface: 0x80 - 0x9F
- OTA MQTT interface: 0xA0 - 0xBF
- OTA HTTP interface: 0xC0 - 0xDF
- OTA PAL inteface: 0xE0 - 0xFF

So OTA agent can define up to 128 different errors, whereas each interface can define up to 32 different errors.


source/include/ota.h
```diff
@@ -264,13 +259,6 @@ typedef struct OtaAgentContext
 /**
  * @constantspage{ota,OTA library}
```
divekarshubham, line 260: Should be moved from constants page to ota_enums

source/include/ota_mqtt_interface.h
```diff
@@ -97,13 +103,13 @@ typedef OtaErr_t ( * OtaMqttUnsubscribe_t )  ( const char * pTopicFilter,
...
  *
- * @return                      OTA_OS_ERR_OK if success , other error code on failure.
+ * @return                      OtaMqttSuccess if success , other error code on failure.
  */
-typedef OtaErr_t ( * OtaMqttPublish_t )( const char * const pacTopic,
-                                         uint16_t usTopicLen,
-                                         const char * pcMsg,
-                                         uint32_t ulMsgSize,
-                                         uint8_t ucQos );
+typedef OtaMqttStatus_t ( * OtaMqttPublish_t )( const char * const pacTopic,
```
divekarshubham, line 108: Should update these parameters according to latest naming conventions

source/ota_http.c
```diff
@@ -49,7 +49,7 @@ static uint32_t currBlock;
  */
 OtaErr_t initFileTransfer_Http( OtaAgentContext_t * pAgentCtx )
 {
-    OtaErr_t status = OTA_ERR_UNINITIALIZED;
+    OtaHttpStatus_t httpErr = OtaHttpSuccess;
```
divekarshubham, line 52: Could be renamed to `httpStatus`, since we may also send success.

source/ota_mqtt.c
```diff
@@ -211,9 +211,9 @@ static uint32_t prvBuildStatusMessageFinish( char * pMsgBuffer,
 /*
  * Subscribe to the OTA job notification topics.
  */
-static OtaErr_t subscribeToJobNotificationTopics( const OtaAgentContext_t * pAgentCtx )
+static OtaMqttStatus_t subscribeToJobNotificationTopics( const OtaAgentContext_t * pAgentCtx )
 {
-    OtaErr_t result = OTA_ERR_UNINITIALIZED;
+    OtaMqttStatus_t mqttErr = OtaMqttSuccess;
```
divekarshubham, line 216: could be renamed to `mqttStatus`

source/ota_mqtt.c
```diff
@@ -653,7 +653,8 @@ OtaErr_t requestJob_Mqtt( OtaAgentContext_t * pAgentCtx )
      * is calculated from the template and the corresponding parameters. */
     char pJobTopic[ TOPIC_GET_NEXT_BUFFER_SIZE ];
     static uint32_t reqCounter = 0;
-    OtaErr_t result = OTA_ERR_UNINITIALIZED;
+    OtaErr_t result = OtaErrRequestJobFailed;
```
divekarshubham, line 656: `result` could be renamed to `otaError` since we are only storing the error codes and not the results of any ops.

source/portable/os/ota_os_freertos.c
```diff
@@ -59,9 +59,9 @@ static void requestTimerCallback( TimerHandle_t T );
 static void selfTestTimerCallback( TimerHandle_t T );
 void ( * timerCallback[ OtaNumOfTimers ] )( TimerHandle_t T ) = { requestTimerCallback, selfTestTimerCallback };
 
-OtaErr_t OtaInitEvent_FreeRTOS( OtaEventContext_t * pEventCtx )
+OtaOsStatus_t OtaInitEvent_FreeRTOS( OtaEventContext_t * pEventCtx )
 {
-    OtaErr_t otaErrRet = OTA_ERR_UNINITIALIZED;
+    OtaOsStatus_t otaErrRet = OtaOsSuccess;
```
divekarshubham, line 64: Maybe change to `otaOsErrRet`?

source/include/ota_mqtt_interface.h
tgsong: let's do that in a separate PR.

source/portable/os/ota_os_freertos.c
tgsong: I'll change to otaOsStatus to align with mqtt and http.

divekarshubham (approved):

aggarg (approved):

PR #95, opened 2020-12-02
Add new event in application callback

pvyawaha:
*Issue #, if available:*

*Description of changes:*
This change adds a new event for the callback from OTA library to application so that application can take further actions for cleanup.


divekarshubham (dismissed):

source/include/ota.h
```diff
@@ -177,8 +178,11 @@ typedef uint32_t OtaErr_t;
  * job has failed in some way and should be rejected.
  *
  * @param[in] eEvent An OTA update event from the OtaJobEvent_t enum.
+ *
+ * @param[in] pData Optional data related to the event.
  */
-typedef void (* OtaAppCallback_t)( OtaJobEvent_t eEvent );
+typedef void (* OtaAppCallback_t)( OtaJobEvent_t eEvent,
+                                   void * pData );
```
dan4thewin, line 185: Why are you using a void * here?

pvyawaha: We have it void ptr so that we application can apply it to any data as the data will be based on the event type.

dan4thewin (approved):

source/ota.c
```diff
@@ -910,7 +913,7 @@ static void dataHandlerCleanup( IngestResult_t result )
     }
 
     /* Let main application know of our result. */
-    otaAgent.OtaAppCallback( ( result == IngestResultFileComplete ) ? OtaJobEventActivate : OtaJobEventFail );
+    otaAgent.OtaAppCallback( ( result == IngestResultFileComplete ) ? OtaJobEventActivate : OtaJobEventFail, NULL );
```
yanjos-dev, line 916: I'd prefer if this was with an if else statement.

But if we want to keep the turnery operator, then can you add another set of parenthesis around the turnery operator?

dan4thewin (approved):

source/ota.c
pvyawaha: I will help in readability and will be added in next PR.

divekarshubham (approved):

PR #98, opened 2020-12-03
Fix PAL error codes

tgsong:
*Issue #, if available:*

*Description of changes:*
OTA PAL error codes are composed inside PAL implementation, thus PAL interface should return uint32_t error codes (combined with main and sub) instead.

**Notes:**

- `otaconfigMAX_NUM_REQUEST_MOMENTUM` is removed from `OtaErrMomentumAbort` because it's directly returned from state machine handler and not reported to the cloud. We will print this error, so composing here is unnecessary and will cause confusing when reading the log error message.
- Default wait time in OTA unit tests is increased to 2s since it seems too short for `test_OTA_ReceiveFileBlockComplete` test.

### DIFF
--- a/source/ota.c
+++ b/source/ota.c
@@ -446,18 +446,18 @@ static OtaErr_t setImageStateWithReason( OtaImageState_t stateToSet,
                                          uint32_t reasonToSet )
 {
     OtaErr_t err = OtaErrNone;
-    OtaPalStatus_t palErr = OtaPalSuccess;
     OtaImageState_t state = stateToSet;
     uint32_t reason = reasonToSet;
+    OtaPalStatus_t palStatus;
 
     /* Call the platform specific code to set the image state. */
-    palErr = otaAgent.pOtaInterface->pal.setPlatformImageState( &( otaAgent.fileContext ), state );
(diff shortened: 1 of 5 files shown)


source/ota.c
```diff
@@ -2311,21 +2312,20 @@ static IngestResult_t ingestDataBlockCleanup( OtaFileContext_t * pFileContext,
         if( pFileContext->pFile != NULL )
         {
             *pCloseResult = otaAgent.pOtaInterface->pal.closeFile( pFileContext );
+            otaPalMainErr = OTA_PAL_MAIN_ERR( *pCloseResult );
+            otaPalSubErr = OTA_PAL_SUB_ERR( *pCloseResult );
```
divekarshubham, line 2316: Can remove this since it is unused, or log it.

tgsong: Which one? `otaPalMainErr` is in comparison, otaPalSubErr in in log on line 2326.

divekarshubham: Ah yes, missed otaPalSubErr in 2326.

divekarshubham (approved):

aggarg (approved):

PR #107, opened 2020-12-06
Eliminate use of snprintf().

gkwicker:
*Issue #, if available:*

*Description of changes:*

The use of `snprintf()` violates Required MISRA rule 21.6.  This change eliminates use of this call from ota_mqtt.c


source/ota_mqtt.c
```diff
@@ -207,6 +215,117 @@ static uint32_t prvBuildStatusMessageFinish( char * pMsgBuffer,
...
+    for( i = 0; strings[ i ] != NULL; i++ )
+    {
+        strncat( pBuffer, strings[ i ], bufferSizeBytes - curLen - 1 );
+        curLen += strlen( strings[ i ] );
+    }
+
+    return curLen;
+}
+
+static size_t stringBuilderUInt32Decimal( char * pBuffer,
```
yanjos-dev, line 268: I see that these functions are only used with static buffers being passed in that are size ( `U32_MAX_LEN` + 1 ), so I don't think we need NULL ptr/size checks, but should we have asserts to make this assumption more obvious?

source/ota_mqtt.c
```diff
@@ -207,6 +215,117 @@ static uint32_t prvBuildStatusMessageFinish( char * pMsgBuffer,
...
+        '8', '9', 'a', 'b',
+        'c', 'd', 'e', 'f',
+    };
+    char workBuf[ U32_MAX_LEN ];
+    char * pCur = workBuf;
+    char * pDest = pBuffer;
+    size_t size = 0;
+    int i;
+
+    /* Render all 8 digits, including leading zeros */
```
yanjos-dev, line 313:
```suggestion
    /* Render all 8 digits, including leading zeros. */
```

source/ota_mqtt.c
```diff
@@ -362,14 +491,25 @@ static OtaMqttStatus_t unsubscribeFromJobNotificationTopic( const OtaAgentContex
...
     assert( pAgentCtx != NULL );
 
     /* Try to unsubscribe from the first of two job topics. */
-    /* coverity[misra_c_2012_rule_21_6_violation] */
-    topicLen = ( uint16_t ) snprintf( pJobTopic,
-                                      sizeof( pJobTopic ),
-                                      pOtaJobsNotifyNextTopicTemplate,
-                                      pAgentCtx->pThingName );
+
+    topicStringParts[ 1 ] = ( const char * ) pAgentCtx->pThingName;
```
yanjos-dev, line 508: Could we use an enum instead of using the index directly? It may make which part of the string is being set. I think this is already kind of obvious due to what is being assigned though, so this is optional.

source/ota_mqtt.c
yanjos-dev: Can you add periods to the rest of the added comments as well for consistency?

yanjos-dev (changes requested):

source/ota_mqtt.c
```diff
@@ -207,6 +215,117 @@ static uint32_t prvBuildStatusMessageFinish( char * pMsgBuffer,
...
+        '8', '9'
+    };
+    char workBuf[ U32_MAX_LEN ];
+    char * pCur = workBuf;
+    char * pDest = pBuffer;
+    size_t size = 0;
+
+    while( value )
+    {
+        *pCur++ = decDigits[ ( value % 10 ) ];
```
cobusve, line 284: What was the reasoning here to not just do *pCur++ = '0' + ( value % 10 );

source/ota_mqtt.c
```diff
@@ -207,6 +215,117 @@ static uint32_t prvBuildStatusMessageFinish( char * pMsgBuffer,
...
+    char workBuf[ U32_MAX_LEN ];
+    char * pCur = workBuf;
+    char * pDest = pBuffer;
+    size_t size = 0;
+    int i;
+
+    /* Render all 8 digits, including leading zeros */
+    for( i = 0; i < 8; i++ )
+    {
+        *pCur++ = hexDigits[ value & 0xf ];
```
cobusve, line 316: I am thinking if we are also doing this for hex digits I would recommend a static string which is file level static and use the same table for both the decimal and the hex lookups, it will save at least 10 bytes - this often matters ...

cobusve (approved):

source/ota_mqtt.c
gkwicker: Good point, will add.

source/ota_mqtt.c
gkwicker: That's what the first version did, but it gives compiler warnings for `int` -> `char` conversions so I changed to this approach.

source/ota_mqtt.c
gkwicker: Yep, will do.

source/ota_mqtt.c
gkwicker: yep i like this, will add.

source/ota_mqtt.c
gkwicker: Let me look into it – it may not make the code any clearer to read, since every topic & payload are slightly different.  I might use `const int` values in the functions themselves, that might be easy to read and more maintainable.

yanjos-dev (approved):

cobusve (approved):

PR #108, opened 2020-12-07
More unit tests coverage

tgsong:
*Issue #, if available:*

*Description of changes:*
- Improve OTA unit tests coverage
- Fix a bug in OTA agent. We should check the size parameter to tell if the buffer is provided by user or not, instead of checking the ptr, because it could be allocated by us previously.

Missing lines are,
- In `parseJobDocFromCustomCallback`, no way to take the true branch because we can't set `otaAgent.customJobCallback` since it's not exposed in our public API. https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/57d01e627c880b33bb08266f24763daa27a410d5/source/ota.c#L1664-L1713
-  In `getFileContextFromJob`. This branch can't happen because this block is under the branch when receiving first job doc, bitmap should either be provided by user or should be NULL (because we're freeing it after complete file transfer). https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/57d01e627c880b33bb08266f24763daa27a410d5/source/ota.c#L2053-L2057


divekarshubham (approved):

test/unit-test/ota_config.h
```diff
@@ -40,6 +40,9 @@
 /* Lower request momentum so that retry fails faster. */
 #define otaconfigMAX_NUM_REQUEST_MOMENTUM       3
 
+/* Make number of blocks per mqtt request larger so we can hit some branch. */
```
gkwicker, line 43: nit: let's reword this for clarity, `/* Use larger number of blocks per mqtt request to increase branch coverage. */`

test/unit-test/ota_utest.c
```diff
@@ -526,8 +562,8 @@ static void otaWaitForEmptyEventWithTimeout( int milliseconds )
         milliseconds--;
     }
 
-    /* Delay for 10 ms to make sure the event handler is finished. */
-    usleep( 10000 );
+    /* Delay for 100 ms to make sure the event handler is finished. */
+    usleep( 100000 );
```
gkwicker, line 566: Why is 100ms the right number?  Why not 1000ms?  Or 10000ms?

gkwicker (approved):
A couple of minor things to consider changing, but overall looks good – approved.

PR #152, opened 2021-02-04
Fix/ota interface coverage

yanjos-dev:
The previous implementation for setting the data interface for OTA had some issues that were preventing branch coverage. It was also missing some test cases. This change:
* Reduces the total number of branches for ota_interface.c
* Makes the string parsing more strict. Only strings that are an exact match with what is expected to be seen in the job document will count as match.
* Adds a macro for a previously hard coded value for the protocol buffer size
* Adds unit tests for covering the potential invalid and valid inputs
* Adds more preprocessor checks for invalid user configurations
* Removed an unnecessary build step from the CI that was causing issues with the coverage report

For testing, I successfully:
* Downloaded OTA updates with MQTT, HTTP, and both MQTT and HTTP enabled.
* Locally tested all permutations of the data interface configurations with the unit tests.

### DIFF
--- a/source/ota.c
+++ b/source/ota.c
@@ -577,7 +577,7 @@ static const char * pOtaEventStrings[ OtaAgentEventMax ] =
 };
 
 static uint8_t pJobNameBuffer[ OTA_JOB_ID_MAX_SIZE ]; /*!< Buffer to store job name. */
-static uint8_t pProtocolBuffer[ 20 ];                 /*!< Buffer to store data protocol. */
+static uint8_t pProtocolBuffer[ OTA_PROTOCOL_BUFFER_SIZE ];                 /*!< Buffer to store data protocol. */
 static Sig256_t sig256Buffer;                         /*!< Buffer to store key file signature. */
 
 static void otaTimerCallback( OtaTimerId_t otaTimerId )
--- a/source/ota.c
+++ b/source/ota.c
@@ -576,9 +576,9 @@ static const char * pOtaEventStrings[ OtaAgentEventMax ] =
     "Shutdown"
 };
 
-static uint8_t pJobNameBuffer[ OTA_JOB_ID_MAX_SIZE ]; /*!< Buffer to store job name. */
-static uint8_t pProtocolBuffer[ OTA_PROTOCOL_BUFFER_SIZE ];                 /*!< Buffer to store data protocol. */
-static Sig256_t sig256Buffer;                         /*!< Buffer to store key file signature. */
+static uint8_t pJobNameBuffer[ OTA_JOB_ID_MAX_SIZE ];       /*!< Buffer to store job name. */
+static uint8_t pProtocolBuffer[ OTA_PROTOCOL_BUFFER_SIZE ]; /*!< Buffer to store data protocol. */
+static Sig256_t sig256Buffer;                               /*!< Buffer to store key file signature. */
 
 static void otaTimerCallback( OtaTimerId_t otaTimerId )
 {
--- a/source/ota_interface.c
+++ b/source/ota_interface.c
@@ -77,16 +77,57 @@ OtaErr_t setDataInterface( OtaDataInterface_t * pDataInterface,
                            const uint8_t * pProtocol )
 {
     OtaErr_t err = OtaErrInvalidDataProtocol;
+    size_t i;
     bool httpInJobDoc;
     bool mqttInJobDoc;
 
-    /* The explicit type casts create additional branches tracked by code
-     * coverage tools that are unreachable. These macros prevent the tools
-     * from tracking branch coverage for these lines. */
-    /* LCOV_EXCL_BR_START */
-    httpInJobDoc = ( strstr( ( const char * ) pProtocol, "HTTP" ) != NULL ) ? true : false;
-    mqttInJobDoc = ( strstr( ( const char * ) pProtocol, "MQTT" ) != NULL ) ? true : false;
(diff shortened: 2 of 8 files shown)


source/ota.c
```diff
@@ -576,9 +576,9 @@ static const char * pOtaEventStrings[ OtaAgentEventMax ] =
     "Shutdown"
 };
 
-static uint8_t pJobNameBuffer[ OTA_JOB_ID_MAX_SIZE ]; /*!< Buffer to store job name. */
-static uint8_t pProtocolBuffer[ 20 ];                 /*!< Buffer to store data protocol. */
-static Sig256_t sig256Buffer;                         /*!< Buffer to store key file signature. */
+static uint8_t pJobNameBuffer[ OTA_JOB_ID_MAX_SIZE ];       /*!< Buffer to store job name. */
+static uint8_t pProtocolBuffer[ OTA_PROTOCOL_BUFFER_SIZE ]; /*!< Buffer to store data protocol. */
```
gkwicker, line 580: Good catch

source/ota_interface.c
```diff
@@ -64,65 +76,87 @@ void setControlInterface( OtaControlInterface_t * pControlInterface )
...
-    OtaErr_t err = OtaErrInvalidDataProtocol;
-    uint32_t i;
-
-    /*
-     * Primary data protocol will be the protocol used for file download if more
-     * than one protocol is selected while creating OTA job.
-     */
-    #if ( configOTA_PRIMARY_DATA_PROTOCOL == OTA_DATA_OVER_MQTT )
-        const char * pProtocolPriority[ OTA_DATA_NUM_PROTOCOLS ] =
+    OtaErr_t err = OtaErrNone;
```
gkwicker, line 79: maintainability: return status values should always be set to error values, e.g.:

```c
OtaErr_t err = OtaErrInvalidDataProtocol;
```

source/ota_interface.c
```diff
@@ -64,65 +76,87 @@ void setControlInterface( OtaControlInterface_t * pControlInterface )
...
+    bool httpInJobDoc;
+    bool mqttInJobDoc;
+
+    /* The explicit type casts create additional branches tracked by code
+     * coverage tools that are unreachable. These macros prevent the tools
+     * from tracking branch coverage for these lines. */
+    /* LCOV_EXCL_BR_START */
+    httpInJobDoc = ( strstr( ( const char * ) pProtocol, "HTTP" ) != NULL ) ? true : false;
```
gkwicker, line 87: Why are we using `strstr()` rather than `strcmp()`?  This will set `httpInJobDoc == true` for `pProtocol="FOOHTTP123";` which doesn't seem correct.  I'd suggest something like:

```c
static const char * pOtaTransferTypes =
{
    /* Do not change the order of these */
    "MQTT",
    "HTTP",
    "MQTTHTTP",
    "HTTPMQTT"
};
static size_t numTransferTypes = sizeof( pOtaTransferTypes ) / sizeof( const char * );
int i;
for( i = 0; i < numTransferTypes; i++ )
{
    if( strcmp( pOtaTransferTypes, ( const char * ) pProtocol ) == 0 )
    {
        break;
    }
}

switch( i )
{
    case 0:   /* MQTT only, you could also use a #define/enum for this value */
         break;
    case 1:
        .
        .
        .
    default:  /* error handling goes here for everything > 3 */
        break;
}
```

nit: also combine assignment with declaration of `httpInJobDoc` & `mqttInJobDoc` since they're immediately assigned.

test/unit-test/ota_utest.c
```diff
@@ -1920,3 +1931,61 @@ void test_OTA_HTTP_strerror( void )
...
+/**
+ * @brief Test that setDataInterface returns an error and does not set the data
+ * interface when provided with an invalid input from a job document.
+ */
+void test_OTA_setDataInterfaceInvalidInput( void )
+{
+    OtaDataInterface_t dataInterface = { NULL, NULL, NULL, NULL };
+    uint8_t pProtocol[ 20 ] = { 0 };
+
+    memcpy( pProtocol, "invalid_protocol", sizeof( "invalid_protocol" ) );
```
gkwicker, line 1985: I'd add test cases with some garbage/unexpected values for `pProtocol` which contain `HTTP` and/or `MQTT` in here.  Those will fail with the `strstr()` implementation above.

gkwicker (changes requested):
See comments

source/ota_interface.c
yanjos-dev: That makes sense. Updated.

source/ota_interface.c
yanjos-dev: This degree of leniency was how the original implementation worked. I didn't want to update the functionality, just the code coverage/implementation. I'm not as concerned about getting strings like "MQTTasdf" due to 1) the data coming from a fixed service that sends consist and 2) the error side effect of parsing that string would still be to set the data interface to MQTT which seems okay to me.

But it makes sense to have consistent and well defined behavior so I'm okay to make this change if you're okay with the behavior of setDataInterface being different from what it was before.

test/unit-test/ota_utest.c
yanjos-dev: Done in this commit
https://github.com/aws/ota-for-aws-iot-embedded-sdk/pull/152/commits/b76cae926d1feacca12a065b071421ec1098a154

gkwicker (approved):

pvyawaha (approved):

PR #153, opened 2021-02-04
clear self-test flag

pvyawaha:
*Issue #, if available:*

*Description of changes:*
The change clears the self-test flag.

### DIFF
--- a/source/ota.c
+++ b/source/ota.c
@@ -774,6 +774,9 @@ static OtaErr_t inSelfTestHandler( const OtaEventData_t * pEventData )
     {
         /* Callback for application specific self-test. */
         otaAgent.OtaAppCallback( OtaJobEventStartTest, NULL );
+
+        /* Clear self-test flag. */
+        otaAgent.fileContext.isInSelfTest = false;
     }
     else
     {
--- a/source/ota.c
+++ b/source/ota.c
@@ -777,6 +777,9 @@ static OtaErr_t inSelfTestHandler( const OtaEventData_t * pEventData )
 
         /* Clear self-test flag. */
         otaAgent.fileContext.isInSelfTest = false;
+
+        /* Stop the self test timer as it is no longer required. */
+        otaAgent.pOtaInterface->os.timer.stop( OtaSelfTestTimer );
     }
     else
     {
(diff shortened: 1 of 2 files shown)


divekarshubham (approved):

source/ota.c
```diff
@@ -774,6 +774,12 @@ static OtaErr_t inSelfTestHandler( const OtaEventData_t * pEventData )
     {
         /* Callback for application specific self-test. */
         otaAgent.OtaAppCallback( OtaJobEventStartTest, NULL );
+
+        /* Clear self-test flag. */
+        otaAgent.fileContext.isInSelfTest = false;
```
gkwicker, line 779: The logic determining whether or not OTA is in self-test is difficult to follow.    Sometimes we call [an internal function `inSelfTest()`](https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/main/source/ota.c#L2318) which in turn [calls a PAL function to see if it is in the "pending commit" state](https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/main/source/ota.c#L623).  In other cases, [we examine this `isInSelfTest` internal member of the file context](https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/main/source/ota.c#L2121).

It also looks like the `isInSelfTest` member of the file context is never set to true in the code, so I'm assuming this value can only come from the service?

Overall comment here – can the logic around the "Self Test" state be simplified?  As it is written it is likely to be difficult to maintain since the state information for "in self test" is located and modified in different places.

gkwicker (changes requested):
Can we simplify the "in self test" logic?  Right now there is no single place where this information (_am I in self-test?_) is stored, and different parts of the code test this condition in different ways.  This type of pattern introduces maintainability risk.

source/ota.c
pvyawaha: This is good point. The main reason for confusion is the inSelfTest function is actually checking the platform is in self test or not. Wheras the flag in the OTA context is the runtime self-test. The inSelfTest function should be renamed to indicate that it is checking platform state platformInSelfTest ?

gkwicker (approved):

divekarshubham (approved):

PR #156, opened 2021-02-10
Improve ota.c line and branch coverage

yanjos-dev:
Add unit tests for state machine handlers in ota.c. This includes changes to:
* Add unit tests for requestJobHandler
* Add unit tests for requestDataHandler
* Add unit tests for initFileHandler
* Add additional mocks
* Add the function declarations of external functions to ota_utest.c

Also includes minor changes to:
* Remove a duplicate void statement in ota.c
* Update some comments and names in ota_utest.c

### DIFF
--- a/test/unit-test/ota_utest.c
+++ b/test/unit-test/ota_utest.c
@@ -343,6 +343,13 @@ OtaErr_t mockMqttInitFileTransferAlwaysFail( OtaAgentContext_t * pAgentCtx )
     return OtaErrInitFileTransferFailed;
 }
 
+OtaErr_t mockMqttInitFileTransferAlwaysSucceed( OtaAgentContext_t * pAgentCtx )
+{
+    ( void ) pAgentCtx;
+
+    return OtaErrNone;
+}
+
 static OtaMqttStatus_t mockMqttPublishAlwaysFail( const char * const unused_1,
                                                   uint16_t unused_2,
                                                   const char * unused_3,
@@ -1967,4 +1974,39 @@ void test_OTA_initFileHandler_TimerFails( void )
     otaInterfaces.os.timer.start = mockOSTimerStartAlwaysFail;
 
     TEST_ASSERT_EQUAL( OtaErrInitFileTransferFailed, initFileHandler( otaEvent.pEventData ) );
+}
+
+/**
+ * @brief Test that initFileHandler returns the proper error when the OS event
+ * send functionality fails.
+ */
+void test_OTA_initFileHandler_EventSendFails( void )
+{
+    OtaEventMsg_t otaEvent = { 0 };
+
+    /* Test failing while trying to send the shutdown event after failing
+     * to initialize the file. */
+    otaInitDefault();
+    /* Fail to initialize the file transfer so the timer is started. */
+    otaDataInterface.initFileTransfer = mockMqttInitFileTransferAlwaysFail;
+
+    /* Simulate reaching the maximum number of attempts before considering
+     * the attempt to be a failure. */
+    otaAgent.requestMomentum = otaconfigMAX_NUM_REQUEST_MOMENTUM;
+    /* Fail to send the OTA event. */
+    otaInterfaces.os.event.send = mockOSEventSendAlwaysFail;
+
+    TEST_ASSERT_EQUAL( OtaErrSignalEventFailed, initFileHandler( otaEvent.pEventData ) );
+
+    /* Test failing while trying to send the request block event after
+     * successfuly initializing the file. */
+    otaInitDefault();
+
+    /* Succeed with the file initialization to then attempt to send the event
+     * for requesting a block. */
+    otaDataInterface.initFileTransfer = mockMqttInitFileTransferAlwaysSucceed;
+    /* Fail to send the OTA event. */
+    otaInterfaces.os.event.send = mockOSEventSendAlwaysFail;
+
+    TEST_ASSERT_EQUAL( OtaErrSignalEventFailed, initFileHandler( otaEvent.pEventData ) );
 }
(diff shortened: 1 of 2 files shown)


test/unit-test/ota_utest.c
```diff
@@ -1935,11 +1982,132 @@ void test_OTA_HTTP_strerror( void )
...
+void test_OTA_initFileHandler_TimerFails( void )
+{
+    OtaEventMsg_t otaEvent = { 0 };
+
+    /* Initialize the OTA interfaces so they are not NULL. */
+    otaInitDefault();
+    /* Fail to initialize the file transfer so the timer is started. */
+    otaDataInterface.initFileTransfer = mockDataInterfaceInitFileTransferAlwaysFail;
+    /* Fail to start the timer. */
+    otaInterfaces.os.timer.start = mockOSTimerStartAlwaysFail;
```
divekarshubham, line 2002: Why do we need to reassign the mocks?

test/unit-test/ota_utest.c
```diff
@@ -1935,11 +1982,132 @@ void test_OTA_HTTP_strerror( void )
...
+
+    /* Test failing while trying to send the request block event after
+     * successfully initializing the file. */
+    otaInitDefault();
+
+    /* Succeed with the file initialization to then attempt to send the event
+     * for requesting a block. */
+    otaDataInterface.initFileTransfer = mockMqttInitFileTransferAlwaysSucceed;
+    /* Fail to send the OTA event. */
+    otaInterfaces.os.event.send = mockOSEventSendAlwaysFail;
```
divekarshubham, line 2037: Duplicate assignment?

test/unit-test/ota_utest.c
```diff
@@ -323,6 +347,29 @@ static OtaMqttStatus_t stubMqttPublish( const char * const unused_1,
...
+
+OtaErr_t mockDataInterfaceInitFileTransferAlwaysFail( OtaAgentContext_t * unused )
+{
+    ( void ) unused;
+
+    return OtaErrInitFileTransferFailed;
+}
+
+
+OtaErr_t mockMqttInitFileTransferAlwaysSucceed( OtaAgentContext_t * unused )
```
divekarshubham, line 366: Can be repurposed for HTTP by renaming it to `mockDataInterfaceInitFileTransfer`

test/unit-test/ota_utest.c
yanjos-dev: I did this so that the two sections of the test aren't dependent on each other. If I was to remove this, and then someone in the future removed the first test or moved it to a different location, it would break this one.  I also wanted to make it clear that it was something that is required to create the scenario that I'm testing.

test/unit-test/ota_utest.c
```diff
@@ -1935,11 +1982,132 @@ void test_OTA_HTTP_strerror( void )
...
+void test_OTA_requestDataHandler_EventSendFails( void )
+{
+    OtaEventMsg_t otaEvent = { 0 };
+
+    otaInitDefault();
+
+    /* File context has a non-zero number of blocks remaining. */
+    otaAgent.fileContext.blocksRemaining = 1U;
+    /* Fail to start the timer. */
+    otaInterfaces.os.timer.start = mockOSTimerStartAlwaysFail;
```
divekarshubham, line 2055: There is a separate TimerFail test for the other functions, maybe fragment this out too?

yanjos-dev: Can you elaborate on what you mean by fragment it out? Do you mean something like have one test that tests that the event handlers cover the scenario where  the request timer fails?

divekarshubham (commented):
You could create a common function like:
```
setupEventFailed(){
    otaInitDefault();
    otaDataInterface.initFileTransfer = mockDataInterfaceInitFileTransferAlwaysFail;
    otaControlInterface.requestJob = mockControlInterfaceRequestJobAlwaysFail;
    otaAgent.requestMomentum = otaconfigMAX_NUM_REQUEST_MOMENTUM;
    otaInterfaces.os.event.send = mockOSEventSendAlwaysFail;
}
```

test/unit-test/ota_utest.c
divekarshubham: Yes like you have with `requestJob`, separate test for timer and event failures in case we want to test out different scenarios in the future.

yanjos-dev: Ah, I understand now. This test is for the event send failing, but something needs to fail for the shutdown event to trigger. I arbitrarily chose the timer to not start as the failure, but thinking on this again I think it's more clear if I set the number of requests to be the max. This is a more obvious scenario where the shutdown event is expected to fail.

test/unit-test/ota_utest.c
yanjos-dev: done

test/unit-test/ota_utest.c
yanjos-dev: Updated this to fail due to making too many requests without success now. I also added some more info to the comment to make it more clear.

yanjos-dev:
> You could create a common function like:
> 
> ```
> setupEventFailed(){
>     otaInitDefault();
>     otaDataInterface.initFileTransfer = mockDataInterfaceInitFileTransferAlwaysFail;
>     otaControlInterface.requestJob = mockControlInterfaceRequestJobAlwaysFail;
>     otaAgent.requestMomentum = otaconfigMAX_NUM_REQUEST_MOMENTUM;
>     otaInterfaces.os.event.send = mockOSEventSendAlwaysFail;
> }
> ```

That would reduce the amount of code in the unit tests, but I'd prefer to be more explicit. This way when someone reads the unit test, it's clear what is failing and what the expected result is.

I also think that in some of the tests certain steps will need to succeed to be able to hit the branch that is under test.

divekarshubham (approved):

aggarg (approved):

PR #157, opened 2021-02-11
Adding logs and version display fix

divekarshubham:
*Issue #, if available:*

*Description of changes:*
* Adding some debugging logs
* Fixing version log to show `major.minor.build format`

### DIFF
--- a/source/ota.c
+++ b/source/ota.c
@@ -1717,6 +1717,8 @@ static DocParseErr_t parseJSONbyModel( const char * pJson,
     const char * pFileParams = NULL;
     uint32_t fileParamsLength = 0;
 
+    LogDebug( ( "JSON received: %s", pJson ) );
+
     /* Fetch the model parameters from the DocModel*/
     pModelParam = pDocModel->pBodyDef;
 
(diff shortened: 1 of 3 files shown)


source/ota.c
```diff
@@ -1854,6 +1856,9 @@ static DocParseErr_t initDocModel( JsonDocModel_t * pDocModel,
 static OtaErr_t validateUpdateVersion( const OtaFileContext_t * pFileContext )
 {
     OtaErr_t err = OtaErrNone;
+    AppVersion32_t newVersion;
```
yanjos-dev, line 1859: Is there a need to create this variable if it's only used for debug printing? In the print statement below, can you use 

"pFileContext->updaterVersion.u.x.minor" instead of "appFirmwareVersion.u.x.minor"?

This would be better than having to void the variable when debug printing is disabled.

divekarshubham: Since `updaterVersion` is only `uint32_t` it has to be first converted to the union `AppVersion32_t` by doing `newVersion.u.unsignedVersion32 = pFileContext->updaterVersion`
without this the formatting doesn't work.

yanjos-dev (approved):

n9wxu (approved):

PR #162, opened 2021-02-15
Removing ota_os_posix.c file from coverage report

divekarshubham:
*Issue #, if available:*

*Description of changes:*


yanjos-dev (approved):

tools/cmock/coverage.cmake
```diff
@@ -69,7 +69,7 @@ execute_process(COMMAND lcov
 # remove source files from dependencies and unit tests
 execute_process(COMMAND lcov
     --rc lcov_branch_coverage=1
-    --remove ${CMAKE_BINARY_DIR}/coverage.info *dependency* *unit-test* /usr* */source/ota.c *CMakeCCompilerId*
+    --remove ${CMAKE_BINARY_DIR}/coverage.info *dependency* *unit-test* /usr* */source/ota.c *CMakeCCompilerId* */source/portable/os/ota_os_posix.c
```
gkwicker, line 72: Why are we removing `ota.c` and `ota_os_posix.c` from coverage checks?

gkwicker (commented):
One question

tools/cmock/coverage.cmake
divekarshubham: For ota.c, we make a copy of it with the static functions stripped off and use that in the unit test, hence we don’t use the source version of the file for the coverage report.
ota_os_posix.c provides abstractions for Linux and not very critical to the coverage report of the library. We aren’t removing these altogether and still have unit tests that assert the basic functionality of these abstractions. The user may or may not use the exact implementation in their code.

aggarg (approved):

aggarg (approved):

yanjos-dev (approved):

PR #163, opened 2021-02-15
Update CI checks to use common GitHub Actions

aggarw13:
Update **ci.yml** to use GitHub Actions from [FreeRTOS/CI-CD-GitHub-Actions](https://github.com/FreeRTOS/CI-CD-Github-Actions) for **coverage-cpp**, **formatting** and **complexity** checks


source/include/ota_config_defaults.h
```diff
@@ -1,343 +1,343 @@
...
- *
- * <b>Default value</b>: Debug logging is turned off, and no code is
- * generated for calls to the macro in the OTA library on compilation.
- */
-#ifndef LogDebug
-    #define LogDebug( message )
-#endif
-
-#endif /* ifndef OTA_CONFIG_DEFAULTS_H */
+/*
```
aggarw13, line 1: This file only has line ending changes from CRLF to LF

yanjos-dev (approved):

.github/workflows/ci.yml
```diff
@@ -68,25 +68,11 @@ jobs:
...
-          if [[ $(echo "$BRANCH_COVERAGE < $MIN_BRANCH_COVERAGE" | bc) -ne 0 ]]; then
-            echo "Branch Coverage is too low."
-            RESULT=1
-          fi
-          exit $RESULT
+        uses: FreeRTOS/CI-CD-Github-Actions/coverage-cop@main
+        with:
+          path: ./build/coverage.info
+          line-min-coverage: 90
+          branch-min-coverage: 70
```
yanjos-dev, line 75: It shows that the branch coverage is set to 70% but in the actual git actions report when I look at the check coverage test, it says the expected amount is 90% and there is some parsing error. Do you know what's causing this?

yanjos-dev (changes requested):
See my comment on the branch coverage.

.github/workflows/ci.yml
aggarw13: Good catch! There was a bug in the action's logic which has been fixed in [PR](https://github.com/FreeRTOS/CI-CD-Github-Actions/pull/11)

yanjos-dev (approved):

aggarg (approved):

PR #167, opened 2021-02-17
Add unit tests for edge cases in the ota.c file

yanjos-dev:
Add unit tests for edge cases in the ota.c file.

**Test failing to send the job notification event**
* Test the scenario where the OTA Agent receives a notification
   for a job document and fails to send the request job document
   event.

**Test edge case for aborting the image state**
* Add a unit test for the scenario where the user calls the API
  to abort the image and the OTA Agent fails to update the job
  status.
* Add a mock for the ota control interface updateJobStatus that
  always fails.

**Test failing to send the event while resuming**
* Add a unit test for the scenario where the OTA state machine is
  suspended and receives an event to resume, attempts to resume,
  and fails to send the resume event.

**Test a timer failure during a block request**
* Add a unit test for the scenario where the OTA Agent is
  requesting a data block and fails to start the request timeout
  timer.

**Add coverage for OTA_EventProcessingTask** 
* Add a unit test to verify that the OTA_EventProcessingTask
  function shutsdown the OTA Agent correctly after receiving an
  event to shutdown.

**Add code coverage for processValidFileContext**
* Add a test for the scenario where we receive a valid job
  document and then fail to send an event for it.

**Add coverage for processNullFileContext** 
* Add a test for failing to send the self-test start task after
  receiving a self test document.


source/ota.c
```diff
@@ -2871,7 +2871,6 @@ static void receiveAndProcessOtaEvent( void )
     }
 }
 
-/* Event Processing loop to run the OTA state machine. */
```
divekarshubham, line 2874: Could we add this/or a better description back?

yanjos-dev: I removed this to be consistent with the rest of the file. The rest of the functions that are in ota.h have a description there (that is more in depth than this comment was) and no description in the ota.c file.

divekarshubham: I would suggest adding comments to the rest of the functions(be it one liners), rather than removing them.

divekarshubham (approved):

dan4thewin (approved):

PR #168, opened 2021-02-17
Fixes for MISRA violations

divekarshubham:
*Issue #, if available:*

*Description of changes:*
Misc fixes for MISRA violations.


source/ota.c
```diff
@@ -2293,8 +2287,17 @@ static OtaFileContext_t * parseJobDoc( const char * pJson,
...
+    {
+        LogInfo( ( "Job parsing sccess: "
+                   "OtaJobParseErr_t=%s, Job name=%s",
+                   OTA_JobParse_strerror( err ), ( const char * ) pFileContext->pJobName ) );
+    }
+    else
+    {
+        /* Handle job parsing error. */
+        handleJobParsingError( pFileContext, err );
+    }
```
yanjos-dev, line 2300: Is adding this branch necessary? It seems to me that the `handleJobParsingError` code already has a case for when `err == OtaJobParseErrNone`. Wouldn't adding this branch make that case in the switch statement redundant/unreachable?

source/ota.c
```diff
@@ -2293,8 +2287,17 @@ static OtaFileContext_t * parseJobDoc( const char * pJson,
         }
     }
 
-    /* Handle job parsing error. */
-    handleJobParsingError( pFileContext, err );
+    if( err == OtaJobParseErrNone )
+    {
+        LogInfo( ( "Job parsing sccess: "
```
yanjos-dev, line 2292:
```suggestion
        LogInfo( ( "Job parsing success: "
```

source/ota.c
```diff
@@ -2832,37 +2835,39 @@ static void receiveAndProcessOtaEvent( void )
...
-
-    /*
-     * Receive the next event form the OTA event queue to process.
-     */
-    if( otaAgent.pOtaInterface->os.event.recv( NULL, &eventMsg, 0 ) == OtaOsSuccess )
+    else
     {
         /*
-         * Search transition index if available in the table.
+         * Receive the next event form the OTA event queue to process.
```
yanjos-dev, line 2841:
```suggestion
         * Receive the next event from the OTA event queue to process.
```

source/ota.c
divekarshubham: Yes, I have removed the case of `OtaJobParseErrNone` from `handleJobParsingError`. This branch is necessary since MISRA requires the error conditions(`err`) to be checked before returning from the function.

yanjos-dev: I see, I missed that. That makes more sense then. My only concern is that if the 'handleJobParsingError' function was called and passed in a no error, error message, then this would cause incorrect behavior. This is okay since this is only being called in this one place but maybe we can either:
1) Update how that default statement works, maybe make it so it's not in the default if that's possible.
2) Just add some information in the description of this function making this assumption clear
3) Add an assert for the input not being a no error, error message.

divekarshubham: Added a comment and an assert in https://github.com/aws/ota-for-aws-iot-embedded-sdk/pull/168/commits/fb39ed24065f4469e15b2bea7e920748901a5c7f

yanjos-dev (dismissed):

dan4thewin (dismissed):

yanjos-dev (approved):

dan4thewin (approved):

PR #171, opened 2021-02-18
Adding tests for improving ota.c coverage

divekarshubham:
*Issue #, if available:*

*Description of changes:*
* Removed unused parameter `OtaJobParseErrBusyWithExistingJob` from the library.
* Added test to check invalid base64 key.
* Added test to check missing required key.
* Added test to check insufficient device and user memory for storing params from job document.


test/unit-test/ota_utest.c
```diff
@@ -1213,6 +1215,35 @@ void test_OTA_ProcessJobDocumentInvalidJson()
     TEST_ASSERT_EQUAL( OtaAgentStateWaitingForJob, OTA_GetState() );
 }
 
+/* Test that the job is rejected if the file key signature cannot be
+ * decoded. */
```
yanjos-dev, line 1219: Want to make these doxygen style comments with @brief to be consistent with the other tests with descriptions? This would apply to the rest of the new tests as well.

test/unit-test/ota_utest.c
```diff
@@ -1643,6 +1674,27 @@ void test_OTA_ReceiveFileBlockCompleteDynamicBufferHttp()
     test_OTA_ReceiveFileBlockCompleteHttp();
 }
 
+/* Test that extractAndStoreArray fails if device does not have sufficient
+ * memory to allocate the string/array. */
+void test_OTA_ExtractArrayMemAllocFails()
+{
+    otaInterfaces.os.mem.malloc = mockMallocAlwaysFail;
+
+    memset( &pOtaAppBuffer, 0, sizeof( pOtaAppBuffer ) );
```
yanjos-dev, line 1683: Are there specific fields in pOtaAppBuffer that you are trying to set to 0? I think changing this line to explicitly set those fields to 0 would be more descriptive towards what you are trying to test.

test/unit-test/ota_utest.c
```diff
@@ -1643,6 +1674,27 @@ void test_OTA_ReceiveFileBlockCompleteDynamicBufferHttp()
...
+    otaGoToState( OtaAgentStateWaitingForFileBlock );
+    TEST_ASSERT_EQUAL( OtaAgentStateWaitingForJob, OTA_GetState() );
+}
+
+/* Test that extractAndStoreArray fails if user buffer is insufficient
+ * to allocate the string/array (here streamname). */
+void test_OTA_ExtractArrayInsufficientBuffer()
+{
+    pOtaAppBuffer.streamNameSize = 5;
```
yanjos-dev, line 1692: Where does the number 5 come from? Can you either:
1) Find a macro (something like min size) and base this number off of that
or
2) Add a comment about why 5? This would help us in the future to see if it needs to be updated based on library changes.

test/unit-test/ota_utest.c
```diff
@@ -1643,6 +1674,27 @@ void test_OTA_ReceiveFileBlockCompleteDynamicBufferHttp()
...
+
+    otaGoToState( OtaAgentStateWaitingForFileBlock );
+    TEST_ASSERT_EQUAL( OtaAgentStateWaitingForJob, OTA_GetState() );
```
yanjos-dev, line 1695: You may want to consider adding a brief description here as to why this is the expected behavior. It's not very clear why going to the `OtaAgentStateWaitingForFileBlock ` state means that we should end up in the `OtaAgentStateWaitingForJob` state. The step in the process of getting to the waiting for file block state and then ending up in the waiting for job state that fails is not clear to me.

yanjos-dev (changes requested):

test/unit-test/ota_utest.c
divekarshubham: Explicitly set streamNameSize to 0.

test/unit-test/ota_utest.c
divekarshubham: Changed to macro OTA_INVALID_STREAM_NAME_SIZE and added comment

test/unit-test/ota_utest.c
divekarshubham: Added description in https://github.com/aws/ota-for-aws-iot-embedded-sdk/pull/171/commits/3f638123d8d8f787b75bdf3e2663b0ee10f9cad5

yanjos-dev (approved):

n9wxu (approved):

yanjos-dev (approved):

PR #175, opened 2021-02-20
Refactoring parseJobDoc and Adding tests for ota.c coverage

divekarshubham:
*Description of changes:*
* Refactored `parseJobDoc` function to be inline with the flow of other functions and make it more testable.
* Moved Job Doc model to the topic with other static arrays
* Added tests for hitting edge conditions and testing private functions in ota.c

### DIFF
--- a/source/ota.c
+++ b/source/ota.c
@@ -2330,11 +2330,7 @@ static OtaFileContext_t * getFileContextFromJob( const char * pRawMsg,
 
     /* Populate an OTA file context from the OTA job document. */
 
-    pUpdateFile = parseJobDoc( otaJobDocModelParamStructure,
-                               OTA_NUM_JOB_PARAMS,
-                               pRawMsg,
-                               messageLength,
-                               &updateJob );
+    pUpdateFile = parseJobDoc( otaJobDocModelParamStructure, OTA_NUM_JOB_PARAMS, pRawMsg, messageLength, &updateJob );
 
     if( updateJob == true )
     {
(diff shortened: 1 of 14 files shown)


source/ota.c
```diff
@@ -285,12 +285,16 @@ static OtaJobParseErr_t validateAndStartJob( OtaFileContext_t * pFileContext,
...
  *
+ * @param[in] pJsonDoc Structure to store the details of keys and where to store them.
+ * @param[in] numJobParams Number of parameters to be extracted.
  * @param[in] pJson JSON job document.
  * @param[in] messageLength Length of the job document.
  * @param[in] pUpdateJob Represents if the job is accepted.
  * @return OtaFileContext_t* File context to store file information.
  */
-static OtaFileContext_t * parseJobDoc( const char * pJson,
+static OtaFileContext_t * parseJobDoc( const JsonDocParam_t * pJsonDoc,
```
yanjos-dev, line 295: the name "pJsonDoc" is misleading compared to the description. This sounds like a pointer to a JSON document. But this is a pointer to a struct that has information about the expected structure of the JSON document. Maybe change it to something like "pJsonDocStructure"?

test/unit-test/ota_utest.c
```diff
@@ -154,6 +160,23 @@ extern OtaErr_t jobNotificationHandler( const OtaEventData_t * pEventData );
 extern OtaErr_t setImageStateWithReason( OtaImageState_t stateToSet,
                                          uint32_t reasonToSet );
 
+/* Static helper function to help initialize the Json document model. */
```
yanjos-dev, line 163: May want  to remove these one line comments and make them more flush like it is with the state machine handler functions.  I would just change the "/*Static helper function under test defined in ota.c. */" to be plural. This way would take up less space and maintain a single source of truth for the description of what the function does.

yanjos-dev (approved):

n9wxu (dismissed):

source/ota.c
divekarshubham: Changed to `pJsonExpectedParams`

test/unit-test/ota_utest.c
divekarshubham: Done

n9wxu (approved):

yanjos-dev (approved):

PR #178, opened 2021-02-22
Fix custom job and no active job logic and logs

pvyawaha:
*Issue #, if available:*

*Description of changes:* This change - 
* Fixes the logic and error codes to handle custom job
* Identifies no active job and removes unnecessary file close
* Removes redundant handling of no active job/parse error
* Cleanup log in cases of custom job parsing failure and no active jobs


divekarshubham (approved):

divekarshubham (dismissed):

divekarshubham (approved):

source/ota.c
```diff
@@ -1941,65 +1930,59 @@ static OtaJobParseErr_t handleCustomJob( const char * pJson,
...
+                                           &jobDoc.jobIdLength,
+                                           NULL ) ) )
+    {
+        /* We have an unknown job parser error. Check to see if we can pass control
+         * to a callback for parsing */
+        otaAgent.OtaAppCallback( OtaJobEventParseCustomJob, &jobDoc );
+
+        if( jobDoc.parseErr == OtaJobParseErrNone )
+        {
+            ( void ) memcpy( otaAgent.pActiveJobName, jobDoc.pJobId, jobDoc.jobIdLength );
```
gkwicker, line 1955: 2 questions:

1. Are we sure that `otaAgent.pActiveJobName` has at least `jobDoc.jobIdLength` bytes of space?  It looks like `otaAgent.pActiveJobName` is `OTA_JOB_ID_MAX_SIZE` bytes based on line 1971.
1. Why are we overwriting the job name with the job ID here?

Some comments explaining this could be helpful.

source/ota.c
```diff
@@ -2231,22 +2214,28 @@ static void handleJobParsingError( const OtaFileContext_t * pFileContext,
                         "OtaJobParseErr_t=%s, Job name=%s",
                         OTA_JobParse_strerror( err ), ( const char * ) pFileContext->pJobName ) );
 
-            /* Assume control of the job name from the context. */
-            ( void ) memcpy( otaAgent.pActiveJobName, pFileContext->pJobName, OTA_JOB_ID_MAX_SIZE );
+            if( strlen( ( const char * ) otaAgent.pActiveJobName ) > 0u )
```
gkwicker, line 2217: A new test was added here, but I don't see any unit test code added for this.  Can we add unit test coverage for this?

source/ota.c
```diff
@@ -1941,65 +1930,59 @@ static OtaJobParseErr_t handleCustomJob( const char * pJson,
...
+                                           strlen( OTA_JSON_JOB_ID_KEY ),
+                                           ( const char ** ) &jobDoc.pJobId,
+                                           &jobDoc.jobIdLength,
+                                           NULL ) ) )
+    {
+        /* We have an unknown job parser error. Check to see if we can pass control
+         * to a callback for parsing */
+        otaAgent.OtaAppCallback( OtaJobEventParseCustomJob, &jobDoc );
+
+        if( jobDoc.parseErr == OtaJobParseErrNone )
```
gkwicker, line 1953: It looks like we're introducing 2 new tests (lines 1953 & 1962), but I don't see unit test cases added for these in this PR.  Can we add them?

gkwicker (changes requested):
See comments

source/ota.c
aggarw13: Regarding 2., it is just copying the job ID (that is parsed from the incoming PUBLISH payload buffer) to the `otaAgent` object so that a status update can be sent for the Job to the service through the `otaControlInterface.updateJobStatus` function.

test/unit-test/ota_utest.c
```diff
@@ -1430,6 +1460,28 @@ void test_OTA_ProcessJobDocumentNumOverflow()
...
+    pOtaJobDoc = JOB_DOC_MISSING_JOB_DOC;
+
+    otaGoToState( OtaAgentStateCreatingFile );
+    TEST_ASSERT_EQUAL( OtaAgentStateWaitingForJob, OTA_GetState() );
+
+    pOtaJobDoc = JOB_DOC_MISSING_JOB_ID;
+    otaGoToState( OtaAgentStateCreatingFile );
+    TEST_ASSERT_EQUAL( OtaAgentStateWaitingForJob, OTA_GetState() );
+}
+
```
yanjos-dev, line 1484: This test is kind of confusing to me. It's called `test_OTA_ProcessCustomJobMissingKeys` but it's using the job documents `JOB_DOC_MISSING_JOB_DOC` and `JOB_DOC_MISSING_JOB_ID`. Based on the name, I would have expected this to use the `JOB_DOC_MISSING_KEY` document.

test/unit-test/ota_utest.c
```diff
@@ -66,6 +66,9 @@
...
 #define JOB_DOC_SERVERFILE_ID             "{\"clientToken\":\"0:testclient\",\"timestamp\":1602795143,\"execution\":{\"jobId\":\"AFR_OTA-testjob20\",\"status\":\"IN_PROGRESS\",\"statusDetails\":{\"self_test\":\"ready\",\"updatedBy\":\"0x1000000\"},\"queuedAt\":1602795128,\"lastUpdatedAt\":1602795128,\"versionNumber\":1,\"executionNumber\":1,\"jobDocument\":{\"afr_ota\":{\"protocols\":[\"MQTT\"],\"streamname\":\"AFR_OTA-XYZ\",\"files\":[{\"filepath\":\"/test/demo\",\"filesize\":" OTA_TEST_FILE_SIZE_STR ",\"fileid\":1,\"certfile\":\"test.crt\",\"sig-sha256-ecdsa\":\"MEQCIF2QDvww1G/kpRGZ8FYvQrok1bSZvXjXefRk7sqNcyPTAiB4dvGt8fozIY5NC0vUDJ2MY42ZERYEcrbwA4n6q7vrBg==\"}] }}}}"
+#define JOB_DOC_MISSING_JOB_DOC           "{\"clientToken\":\"0:testclient\",\"timestamp\":1602795143,\"execution\":{\"jobId\":\"AFR_OTA-testjob20\",\"status\":\"QUEUED\",\"queuedAt\":1602795128,\"lastUpdatedAt\":1602795128,\"versionNumber\":1,\"executionNumber\":1}}"
```
yanjos-dev, line 69: Can you elaborate on what this job document represents? It's confusing to me that it's a job document that's missing a job document.

divekarshubham: Each json document for an update has a key called 'jobDocument' which contains actual information about the update parameters. In this scenario this key is missing.

test/unit-test/ota_utest.c
```diff
@@ -2229,6 +2281,33 @@ void test_OTA_RefreshWithSameJobDoc()
...
+
+    otaGoToState( OtaAgentStateWaitingForFileBlock );
+    TEST_ASSERT_EQUAL( OtaAgentStateWaitingForFileBlock, OTA_GetState() );
+
+    /* Set the event send interface to a mock function that allows events to be sent continuously.
+     * We need this to go through the process of refreshing job doc. */
+    otaInterfaces.os.event.send = mockOSEventSend;
+
+    /* First send request job doc event while we're in progress, this should make OTA agent to
+     * to request job doc again and transit to waiting for job state. */
```
yanjos-dev, line 2298: I recommend updating this comment to something like this:
"Send the request job document event while the OTA Agent is already in the process of downloading a file. The OTA Agent should cancel the current job and begin waiting for the next job document."

test/unit-test/ota_utest.c
divekarshubham: For a custom job, we have a check for only 2 keys: `jobid` and `jobdocument`. Other job  parsing tests are prefixed with 'ProcessJobDocument' to show the distinction.

test/unit-test/ota_utest.c
yanjos-dev: I see, that makes sense.

source/ota.c
pvyawaha: Added check in fd787aa3109f0ec2a55a00771bb56738580a24a3

gkwicker (changes requested):
Let's add a unit test for the new length check if we can.

yanjos-dev (approved):

pvyawaha:
> Let's add a unit test for the new length check if we can.

Added unit tests in the commit https://github.com/aws/ota-for-aws-iot-embedded-sdk/pull/178/commits/1443c4d5e7978945ae5f08d8fc08a77997444c93

gkwicker (approved):

PR #179, opened 2021-02-23
Add unit tests for ota.c

yanjos-dev:
* Add tests for the PAL failing while receiving a block.
* Test ingestDataBlockCleanup with a NULL file.
* Fix the downgrade test.
* Add a test for when the new version of the software is the same as what is currently running.
* Test receiving a job doc with a non-default file type.
* Remove unreachable code from ota.c.
* Test NULL inputs to verifyActiveJobStatus


gkwicker (approved):

test/unit-test/ota_utest.c
```diff
@@ -2048,6 +2125,33 @@ void test_OTA_SelfTestJobDowngrade()
...
+
+    otaGoToState( OtaAgentStateWaitingForJob );
+    TEST_ASSERT_EQUAL( OtaAgentStateWaitingForJob, OTA_GetState() );
+
+    otaReceiveJobDocument();
+    receiveAndProcessOtaEvent();
+    TEST_ASSERT_EQUAL( true, resetCalled );
+}
+
+void test_OTA_SelfTestJobFileType()
```
divekarshubham, line 2140: Might want to rename to reflect the condition tested?  `test_OTA_SelfTestJobNonFirmwareFileType`? or add a comment with the scenario.

test/unit-test/ota_utest.c
```diff
@@ -1727,6 +1755,55 @@ void test_OTA_ReceiveLastFileBlockTooSmall()
     TEST_ASSERT_EQUAL( OtaAgentStateWaitingForJob, OTA_GetState() );
 }
 
+void test_OTA_ReceiveWithNullFile()
```
divekarshubham, line 1758: could you add a comment with what happens when you receive a null file pointer?

divekarshubham (approved):

test/unit-test/ota_utest.c
yanjos-dev: done

test/unit-test/ota_utest.c
yanjos-dev: done

gkwicker (approved):

Issue #210, opened 2021-04-23
BUG: Misleading documentation about OtaAppCallback_t


b1ackviking:
https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/ed30b37d302bbbea11866ba8d445a9da0f85bd9b/source/include/ota.h#L213-L219

The documentation states that the application callback is optional, while it is not. The application must provide a valid function pointer to the OTA_Init function. The callback pointer is [stored in the agent context without any check](https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/main/source/ota.c#L3103) and called everywhere without checks. There is no default callback.

PR #211, opened 2021-04-23
Fix misleading documentation of the OtaAppCallback_t

b1ackviking:
Resolves: https://github.com/aws/ota-for-aws-iot-embedded-sdk/issues/210

Remove the false statement that the application callback is optional. The application callback is required and never checked for NULL. There is also no default callback in the code.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of your choice.

### DIFF
--- a/source/include/ota.h
+++ b/source/include/ota.h
@@ -210,7 +210,7 @@ typedef struct OtaJobDocument
  * @ingroup ota_callback_types
  * @brief OTA update complete callback function typedef.
  *
- * The user has to register a callback function when initializing the OTA Agent. This
+ * The user must register a callback function when initializing the OTA Agent. This
  * callback is used to notify the main application when the OTA update job is complete.
  * Typically, it is used to reset the device after a successful update by calling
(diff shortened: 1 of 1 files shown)


pvyawaha (dismissed):

source/include/ota.h
```diff
@@ -210,13 +210,11 @@ typedef struct OtaJobDocument
  * @ingroup ota_callback_types
  * @brief OTA update complete callback function typedef.
  *
- * The user may register a callback function when initializing the OTA Agent. This
+ * The user has to register a callback function when initializing the OTA Agent. This
```
n9wxu, line 213: I would prefer "must" to "has to"

n9wxu (commented):
One minor change

pvyawaha (approved):

source/include/ota.h
pvyawaha: Fixed in 8ed3d1fabbd5ecd9cd9e602491554bd7c37259b4

n9wxu (approved):

PR #215, opened 2021-05-10
Fix ignored zero version number in job status

boraozgen:
*Issue #, if available:*

*Description of changes:*

Zeroes in version numbers were being ignored when reporting job status.


boraozgen:
Of course, it does not work :) another approach is needed to print the zeroes.

PR #225, opened 2021-05-26
Move app callback after populating the jobDoc var

pvyawaha:
*Description of changes:* The app callback should be moved after populating the contents in jobDoc so that it is available as data to application.


divekarshubham (approved):

paulbartell (approved):
Looks relatively straight forward. Please add some unit test cases when possible that will catch this class of bug.

PR #227, opened 2021-07-13
CBMC Proof for OTA library

Bujain:
Added CBMC Proofs for Protocol interface and HTTP interface. 


test/cbmc/negative_tests/README.md
```diff
@@ -0,0 +1 @@
+../aws-templates-for-cbmc-proofs/template-for-repository/negative_tests/README.md
```
yanjos-dev, line 1: What is the purpose of the `test/cbmc/negative_tests` directory?

test/cbmc/include/README.md
```diff
@@ -0,0 +1 @@
+../aws-templates-for-cbmc-proofs/template-for-repository/include/README.md
```
yanjos-dev, line 1: This README file seems like the only file in this directory. What is the purpose of the directory? I see that this is the filepath to the submodule, but it may be good to add a description here as well. Also, if we aren't going to require this directory, then we should remove it.

yanjos-dev (commented):
Is there a reason to group the proofs? It doesn't seem like any of the other libraries are doing this. IMO it makes sense to keep them individual since we aren't testing groups of functionalities, but the memory safety of individual functions.

source/ota_mqtt.c
```diff
@@ -1193,7 +1193,7 @@ OtaErr_t cleanupControl_Mqtt( const OtaAgentContext_t * pAgentCtx )
             result = OtaErrCleanupControlFailed;
         }
     }
-
```
yanjos-dev, line 1196: This seems like it was added on accident. Can you remove this extra whitespace?

source/ota_mqtt.c
```diff
@@ -1251,6 +1251,6 @@ const char * OTA_MQTT_strerror( OtaMqttStatus_t status )
             str = "InvalidErrorCode";
             break;
     }
-
+
```
yanjos-dev, line 1254: This seems like it was added on accident. Can you remove this extra whitespace?

test/cbmc/proofs/Makefile-project-defines
```diff
@@ -0,0 +1,38 @@
...
+LINK_FLAGS = -std=gnu90
+
+# Preprocessor include paths -I...
+# Consider adding
+#     INCLUDES += -I$(CBMC_ROOT)/include
+# You will want to decide what order that comes in relative to the other
+# include directories in your project.
+#
+INCLUDES += -I$(SRCDIR)/source/include
+INCLUDES += -I$(SRCDIR)/test/unit-test
```
yanjos-dev, line 29: Are these paths correct? What is SRCDIR being set to? Is this the root of the directory? I'm wondering why we would include the unit-test directory. It seems like this would not be relevant.

test/cbmc/proofs/Makefile-project-defines
```diff
@@ -0,0 +1,38 @@
...
+
+# Preprocessor definitions -D...
+DEFINES = -Dhttp_EXPORTS
```
yanjos-dev, line 32: What is this define?

test/cbmc/proofs/Makefile-project-defines
```diff
@@ -0,0 +1,38 @@
...
+# Use this file to give project-specific definitions of the command
+# line arguments to pass to CBMC tools like goto-cc to build the goto
+# binaries and cbmc to do the property and coverage checking.
+#
+# Use this file to override most default definitions of variables in
+# Makefile.common.
+################################################################
+
+# Flags to pass to goto-cc for compilation (typically those passed to gcc -c)
+COMPILE_FLAGS = -fPIC
```
yanjos-dev, line 17: What is this flag doing?

test/cbmc/proofs/Makefile-template-defines
```diff
@@ -0,0 +1,19 @@
...
+LITANI ?= $(abspath $(PROOF_ROOT)/../litani/litani)
+
+
+# Name of this proof project, displayed in proof reports. For example,
+# "s2n" or "Amazon FreeRTOS". For projects with multiple proof roots,
+# this may be overridden on the command-line to Make, for example
+#
+# 	  make PROJECT_NAME="FreeRTOS MQTT" report
+#
+PROJECT_NAME = "FreeRTOS OTA"
```
yanjos-dev, line 18: Where did ou get this project name from? The full name of the library is:
"AWS IoT Over-the-air Update Library". We may want to change it to this.

test/cbmc/proofs/ota_http/OTA_HTTP_strerror/Makefile
```diff
@@ -0,0 +1,26 @@
...
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+PROJECT_SOURCES += $(SRCDIR)/source/ota_http.c
+
+# If this proof is found to consume huge amounts of RAM, you can set the
+# EXPENSIVE variable. With new enough versions of the proof tools, this will
+# restrict the number of EXPENSIVE CBMC jobs running at once. See the
+# documentation in Makefile.common under the "Job Pools" heading for details.
+# EXPENSIVE = true
+
+include ../../Makefile.common
```
yanjos-dev, line 26: Wouldn't this be the `root/test/cbmc/proofs/ota_http/` directory? I don't see a "Makefile.common" file there.

test/cbmc/proofs/ota_http/OTA_HTTP_strerror/OTA_HTTP_strerror_harness.c
```diff
@@ -0,0 +1,36 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * @file OTA_HTTP_strerror_harness.c
+ * @brief Implements the proof harness for OTA_HTTP_strerror function.
+ */
+
+#include "ota_http_private.h"
+#include "ota_http_interface.h"
```
yanjos-dev, line 29: It would be good to comment to provide some information with why these are being included/why they are needed. Apply to rest as well.

yanjos-dev (commented):
Is it possible to add the CBMC proofs to the github actions tests? This way they would run on every PR and we can know when we make a breaking change. This will avoid some last second release scrambling.

test/cbmc/proofs/ota_http/OTA_HTTP_strerror/OTA_HTTP_strerror_harness.c
```diff
@@ -0,0 +1,36 @@
...
+
+void OTA_HTTP_strerror_harness()
+{
+  OtaHttpStatus_t status;
+
+  OTA_HTTP_strerror(status);
```
yanjos-dev, line 35: This function returns a string. I don't think that we need to use it, but can you cast the return to void so it's not an unused return statement?

test/cbmc/proofs/ota_http/OTA_HTTP_strerror/README.md
```diff
@@ -0,0 +1,20 @@
+OTA_HTTP_strerror proof
+==============
+
+This directory contains a memory safety proof for OTA_HTTP_strerror .
```
yanjos-dev, line 4:
```suggestion
This directory contains a memory safety proof for OTA_HTTP_strerror.
```

test/cbmc/proofs/ota_http/OTA_HTTP_strerror/cbmc-proof.txt
```diff
@@ -0,0 +1 @@
+# This file marks this directory as containing a CBMC proof.
```
yanjos-dev, line 1: It would be nice to add some more information about why this needs to be here so it is not removed in the future. I see that it marks the directory, but for what tool/script?

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
```diff
@@ -0,0 +1,60 @@
...
+
+  uint8_t* pMessageBuffer;
+  size_t messageSize;
+  int32_t* pFileId;
+  int32_t* pBlockId;
+  int32_t* pBlockSize;
+  uint8_t** pPayload;
+  size_t * pPayloadSize;
+
+  __CPROVER_assume(messageSize < (OTA_FILE_BLOCK_SIZE + 1) && messageSize != 0);
```
yanjos-dev, line 41: Can you add a comment about this assumption? Ideally we don't have any assume statements in our proofs. So if we do have an assert, we should comment about why this is a safe assumption and why we are not updating the code to make it so that we don't need the assumption. Apply this to similar statements please.

yanjos-dev:
It looks like the spell check and formatting check are both failing. Can you update the PR to fix these?

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
```diff
@@ -0,0 +1,60 @@
...
+ * @file decodeFileBlock_Http_harness.c
+ * @brief Implements the proof harness for decodeFileBlock_Http function.
+ */
+
+#include "ota_http_private.h"
+
+
+void decodeFileBlock_Http_harness()
+{
+
```
AniruddhaKanhere, line 32: Remove extra line.

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
AniruddhaKanhere, line 41: The formatting seems off. Have we run this through uncrustify? For example, this line should be:
```suggestion
  __CPROVER_assume( ( messageSize < ( OTA_FILE_BLOCK_SIZE + 1 ) ) && ( messageSize != 0 ) );

```

Also, using parenthesis to separate expressions is a good practice.

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
```diff
@@ -0,0 +1,60 @@
...
+  size_t messageSize;
+  int32_t* pFileId;
+  int32_t* pBlockId;
+  int32_t* pBlockSize;
+  uint8_t** pPayload;
+  size_t * pPayloadSize;
+
+  __CPROVER_assume(messageSize < (OTA_FILE_BLOCK_SIZE + 1) && messageSize != 0);
+
+  pMessageBuffer = (uint8_t*)malloc(messageSize);
```
AniruddhaKanhere, line 43: Are we using malloc-can-fail flag here? If not, then it might be better to use safeMalloc() - which would be something like below and would mimic the operation of the flag. (I recommend using the flag)
```
void * Safemalloc( size_t xWantedSize )
{
    if( nondet_bool() )
    {
        return malloc( xWantedSize );
    }
    else
    {
        return NULL;
    }
}
```

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
```diff
@@ -0,0 +1,60 @@
...
+
+  pFileId = (int32_t *)malloc(sizeof(int32_t));
+  pBlockId = (int32_t *)malloc(sizeof(int32_t));
+  pBlockSize = (int32_t *)malloc(sizeof(int32_t));
+
+  char* temp = malloc(messageSize);
```
AniruddhaKanhere, line 49: Can we rename `temp` to something more descriptive?

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
```diff
@@ -0,0 +1,60 @@
...
+  __CPROVER_assume(temp != NULL);
+
+  pPayload = &temp;
+
+  pPayloadSize = (size_t *)malloc(sizeof(size_t));
+
+  __CPROVER_assume(pMessageBuffer != NULL && pFileId != NULL && pBlockId != NULL &&
```
AniruddhaKanhere, line 56: While this is a perfectly good way to write this, but I personally prefer these assumptions to be separated. That seems cleaner and we can comment the reason behind each assumption. Like so:
```
/* Comment why we need to assume this. */
__CPROVER_assume( ConditionA != FLASE );

/* Comment why we need to assume this. */
__CPROVER_assume( ConditionB == FLASE );
```

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
```diff
@@ -0,0 +1,60 @@
...
+            pBlockSize != NULL && pPayloadSize != NULL);
+  
+  decodeFileBlock_Http(pMessageBuffer,messageSize,pFileId,pBlockId,pBlockSize,pPayload,pPayloadSize);
```
AniruddhaKanhere, line 59: Is there anything to assert on after the call to this function? If not, ignore the comment.

test/cbmc/proofs/ota_http/initFileTransfer_Http/initFileTransfer_Http_harness.c
```diff
@@ -0,0 +1,54 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * @file initFileTransfer_Http_harness.c
+ * @brief Implements the proof harness for initFileTransfer_Http function.
+ */
+#include "ota_http_private.h"
+
+OtaHttpStatus_t init(char* pUrl){
```
AniruddhaKanhere, line 29: This may warrant a comment - why is this function declared here. Is this a stub?

test/cbmc/proofs/ota_http/initFileTransfer_Http/initFileTransfer_Http_harness.c
```diff
@@ -0,0 +1,54 @@
...
+  OtaHttpStatus_t status;
```
AniruddhaKanhere, line 30: Also, do we need to assert that pUrl is not-NULL? `__CPROVER_assert( pUrl != NULL, "URL cannot be NULL" );`

test/cbmc/proofs/ota_http/requestDataBlock_Http/requestDataBlock_Http_harness.c
```diff
@@ -0,0 +1,55 @@
...
+{
+  OtaAgentContext_t *pAgentCtx;
+  OtaFileContext_t fileContext; 
+  OtaHttpInterface_t http;
+
+  OtaInterfaces_t interface; 
+
+  pAgentCtx = (OtaAgentContext_t*) malloc(sizeof(OtaAgentContext_t));
+
+  __CPROVER_assume(pAgentCtx != NULL);
```
AniruddhaKanhere, line 45: I would also suggest adding comments on assumptions across the proofs such that the reader can know why was the assumption made without going through the code.

AniruddhaKanhere (changes requested):
Thank you @Bujain for this PR! Overall the proofs look good. But there are some changes that I would like to request. See the comments.

test/cbmc/negative_tests/conversion_check/conversion_check_harness.c
```diff
@@ -0,0 +1 @@
+../../aws-templates-for-cbmc-proofs/template-for-repository/negative_tests/conversion_check/conversion_check_harness.c
```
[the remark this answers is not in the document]
Bujain: Yes, it is auto-generated.

.gitmodules
```diff
@@ -8,3 +8,9 @@
 	path = test/unit-test/CMock
 	url = https://github.com/ThrowTheSwitch/CMock.git
 	update = none
+[submodule "test/cbmc/aws-templates-for-cbmc-proofs"]
+	path = test/cbmc/aws-templates-for-cbmc-proofs
+	url = https://github.com/awslabs/aws-templates-for-cbmc-proofs.git
```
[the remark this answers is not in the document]
Bujain: I have added the update line to the commit

test/cbmc/include/README.md
Bujain: This include directory is only used to store any header files which are required by the test harness file. Since there are no header files included as of now, I have removed the following directory

test/cbmc/proofs/Makefile-project-defines
Bujain: Yes, the SRCDIR is set to the root of the directory. I'm not sure why the unit-test was included too. I will remove the include line if it is not required.

test/cbmc/proofs/Makefile-template-defines
Bujain: When I was working on the initial proofs, I might have added it without asking anyone. I have changed the PROJECT_NAME now

test/cbmc/proofs/ota_http/OTA_HTTP_strerror/Makefile
Bujain: No. Since the makefile is in `root/test/cbmc/proofs/ota_http/OTA_HTTP_strerror/ ` it would go back to `root/test/cbmc/proofs/ota_http/`

test/cbmc/negative_tests/README.md
Bujain: Right now the CBMC proof does not require negative tests. I have removed the directory.

AniruddhaKanhere:
@Bujain, would you mind adding the word `setcontrolinterface` (all lower case) to the file here: https://github.com/Bujain/ota-for-aws-iot-embedded-sdk/blob/ota_proof/tools/lexicon.txt so that the spell check can pass?

Thank you.

Bujain:
> 
> 
> @Bujain, would you mind adding the word `setcontrolinterface` (all lower case) to the file here: https://github.com/Bujain/ota-for-aws-iot-embedded-sdk/blob/ota_proof/tools/lexicon.txt so that the spell check can pass?
> 
> Thank you.

I've changed the lexicon file and now it passes the spellcheck

test/cbmc/proofs/ota_http/OTA_HTTP_strerror/Makefile
yanjos-dev: I see, you are correct. I checked that file and it seems like it's just defined to a path. I went to that path and the example common makefile looked quite different. Is this Makefile.common file supposed to be updated?

Bujain: It's not supposed to be updated but it might have changed from the previous version of the template. I'll look into it.

Bujain:
> 
> 
> It looks like the spell check and formatting check are both failing. Can you update the PR to fix these?

I have updated the PR to fix this.

test/cbmc/proofs/Makefile-project-defines
Bujain: There are some header files which are used by the ota library present in the unit-test directory.

test/cbmc/proofs/Makefile-project-defines
Bujain: I have updated the makefile and removed unnecessary flags

test/cbmc/proofs/ota_http/OTA_HTTP_strerror/Makefile
Bujain: There seem to no changes to this `Makefile,common` from the example `Makefile.common` .

test/cbmc/proofs/ota_http/OTA_HTTP_strerror/cbmc-proof.txt
Bujain: This file is required by the `run-cbmc-proofs.py` to differentiate between a proof directory and a normal directory. I have added the comment in the last commit.

test/cbmc/proofs/Makefile-project-defines
Bujain: see previous comment.

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
Bujain: I have removed this assumption since it was wrong and did not make any difference to the proof.

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
Bujain: Removed it.

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
Bujain: I have removed this assumption since it was not making difference and also have uncrustified it.

test/cbmc/proofs/Makefile-project-defines
yanjos-dev: What files are these? If they are things like the ota_config.h header file, then we should have a unique copy of this file for CBMC proofs. This config file is only needed though if the default config files are not sufficient.

Bujain: Yes it's the config file.

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
Bujain: We are using the malloc can fail flag in the makefile.common file.

test/cbmc/proofs/ota_http/OTA_HTTP_strerror/OTA_HTTP_strerror_harness.c
```diff
@@ -0,0 +1,42 @@
...
+void OTA_HTTP_strerror_harness()
+{
+    OtaHttpStatus_t status;
+    const char * pstr = NULL;
+
+    /* Call function under test. */
+    pstr = OTA_HTTP_strerror( status );
+
+    /* Assert to check the return of the function is not null. If it is,
+     *  then there is some problem. */
```
AniruddhaKanhere, line 40: We might want to reword the comment here to explain WHY the pointer can be non-NULL.

test/cbmc/proofs/ota_http/cleanupData_Http/cleanupData_Http_harness.c
```diff
@@ -0,0 +1,67 @@
...
+    OtaAgentContext_t * pAgentCtx;
+    OtaHttpInterface_t pHttp;
+    OtaInterfaces_t * pInterfaces;
+    OtaHttpStatus_t status;
+
+    pAgentCtx = ( OtaAgentContext_t * ) malloc( sizeof( OtaAgentContext_t ) );
+    pInterfaces = ( OtaInterfaces_t * ) malloc( sizeof( OtaInterfaces_t ) );
+
+    /* The funciton requires Agent and the interface to be initialized and thus
+     * they can't be NULL. If they are, they will hit an assert in the function. */
```
AniruddhaKanhere, line 50: Awesome! I like this comment. It checks all the boxes :)

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
```diff
@@ -0,0 +1,90 @@
...
+    pMessageBuffer = ( uint8_t * ) malloc( messageSize );
+
+    /* The size of the message buffer cannot be null. */
+    __CPROVER_assume( size != NULL );
+
+    /* Initializing the variable. */
+    pPayload = &size;
+    pPayloadSize = &payloadSize;
+
+    /* The size of message buffer is not null. */
```
AniruddhaKanhere, line 64: These comments should explain why this should be NULL. Would the function hit an assert? Or is the pointer dereferenced since it was checked before in the call tree?

test/cbmc/proofs/ota_http/initFileTransfer_Http/initFileTransfer_Http_harness.c
```diff
@@ -0,0 +1,69 @@
...
+    OtaAgentContext_t * pAgentCtx;
+    OtaFileContext_t fileContext;
+    OtaHttpInterface_t http;
+    OtaHttpStatus_t status;
+    OtaInterfaces_t interface;
+
+    pAgentCtx = ( OtaAgentContext_t * ) malloc( sizeof( OtaAgentContext_t ) );
+
+    /* The funciton requires Agent and the interface to be initialized and
+     *  thus they can't be NULL. If they are, they will hit an assert in the function. */
```
AniruddhaKanhere, line 49: Nice job with the comment!

test/cbmc/proofs/ota_mqtt/OTA_MQTT_strerror/OTA_MQTT_strerror_harness.c
```diff
@@ -0,0 +1,40 @@
...
+
+void OTA_MQTT_strerror_harness()
+{
+    OtaMqttStatus_t status;
+    const char * pstr;
+
+    /* Call function under test. */
+    pstr = OTA_MQTT_strerror( status );
+
+    /* Return from the function cannot be NULL. */
```
AniruddhaKanhere, line 38: We might want to reword this comment to say WHY can this be non-NULL.

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
Bujain: Have added the assert statement.

test/cbmc/proofs/ota_http/initFileTransfer_Http/initFileTransfer_Http_harness.c
Bujain: Yes it is a stub. I have added appropriate comments to the code

test/cbmc/proofs/ota_http/cleanupData_Http/cleanupData_Http_harness.c
```diff
@@ -0,0 +1,67 @@
...
+    /* The function requires Agent and the interface to be initialized and thus
+     * they can't be NULL. If they are, they will hit an assert in the function. */
+    __CPROVER_assume( pAgentCtx != NULL && pInterfaces != NULL );
+
+    /* Updating the function pointer in pHttp to the stub. */
+    pHttp.deinit = deinit;
+
+    /* Update the interface and the Agent. */
+    pInterfaces->http = pHttp;
+    pAgentCtx->pOtaInterface = pInterfaces;
```
yanjos-dev, line 58: This assignment is making it so that the pHttp.deinit pointer is always set to a valid function when the CBMC proof harness calls cleanupData_Http().

Do we know that this will always be the case when the code is running? Assignments like this are similar to the CBMC assert/assume statements where it's constraining the input.

If this is an agreement in the code that cleanupData_Http will always have valid pointers in the input, then we should document how we know this and why it's needed.

test/cbmc/proofs/Makefile-project-defines
```diff
@@ -0,0 +1,37 @@
...
+# Flags to pass to goto-cc for linking (typically those passed to gcc)
+#LINK_FLAGS = -std=gnu90
+
+# Preprocessor include paths -I...
+# Consider adding
+#     INCLUDES += -I$(CBMC_ROOT)/include
+# You will want to decide what order that comes in relative to the other
+# include directories in your project.
+#
+INCLUDES += -I$(SRCDIR)/source/include
```
yanjos-dev, line 28: I noticed that you also have this include path being appended to the path in the individual proof harnesses. Would this be adding the same path to the include path twice?

If so, we should consider removing this line or removing the line from the individual proofs.

test/cbmc/proofs/ota_http/initFileTransfer_Http/initFileTransfer_Http_harness.c
```diff
@@ -0,0 +1,69 @@
...
+    OtaFileContext_t fileContext;
+    OtaHttpInterface_t http;
+    OtaHttpStatus_t status;
+    OtaInterfaces_t interface;
+
+    pAgentCtx = ( OtaAgentContext_t * ) malloc( sizeof( OtaAgentContext_t ) );
+
+    /* The function requires Agent and the interface to be initialized and
+     *  thus they can't be NULL. If they are, they will hit an assert in the function. */
+    __CPROVER_assume( pAgentCtx != NULL );
```
yanjos-dev, line 50: My understanding is that __CPROVER_assume is effectively telling the CBMC tools to never assign pAgentCtx to NULL here.

Does this statement have any effect if we are already assigning a value to pAgentCtx? I believe that if you directly assign a value to a variable, then the CBMC tools will not alter the value of that variable.

test/cbmc/proofs/ota_http/requestDataBlock_Http/requestDataBlock_Http_harness.c
```diff
@@ -0,0 +1,72 @@
...
+void requestDataBlock_Http_harness()
+{
+    OtaAgentContext_t * pAgentCtx;
+    OtaFileContext_t fileContext;
+    OtaHttpInterface_t http;
+    OtaInterfaces_t interface;
+    OtaErr_t err;
+
+    /* Allocating memory to the agent context. */
+    pAgentCtx = ( OtaAgentContext_t * ) malloc( sizeof( OtaAgentContext_t ) );
```
yanjos-dev, line 49: Same feedback as the previous proof. Please see about constraining this less and free this memory.

test/cbmc/proofs/ota_interface/setControlInterface/setControlInterface_harness.c
```diff
@@ -0,0 +1,40 @@
...
+void setControlInterface_harness()
+{
+    OtaControlInterface_t * pControlInterface;
+    OtaControlInterface_t controlInterface;
+
+    /* Initialize the control interface. */
+    pControlInterface = &controlInterface;
+
+    /* Call to the function under test. */
+    setControlInterface( pControlInterface );
```
yanjos-dev, line 39: See my previous comment, it would be good to have the proof check for the value being NULL, and the value being set to an unbounded variable.

test/cbmc/proofs/ota_interface/setDataInterface/setDataInterface_harness.c
```diff
@@ -0,0 +1,53 @@
...
+{
+    OtaDataInterface_t * pDataInterface;
+    OtaDataInterface_t dataInterface;
+    OtaErr_t err;
+
+    uint8_t * pProtocol;
+    size_t size;
+
+    /* Initialize the variable. */
+    pProtocol = ( uint8_t * ) malloc( size );
```
yanjos-dev, line 40: How does a malloc with a variable that's not set work with these proofs? Is there an upper limit on what size could be? If the value can be the max size of an int, this could cause some issues running the test.

test/cbmc/proofs/ota_mqtt/OTA_MQTT_strerror/OTA_MQTT_strerror_harness.c
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+#include "ota_mqtt_private.h"
+
+void OTA_MQTT_strerror_harness()
+{
+    OtaMqttStatus_t status;
+    const char * pstr;
+
+    /* Call function under test. */
```
yanjos-dev, line 35: Nitpick: This comment can be removed. I think it will be clear to the reader that the function under test is being called without this.

yanjos-dev (changes requested):

test/cbmc/proofs/ota_interface/setDataInterface/setDataInterface_harness.c
Bujain: Yes, there is a upper limit on the `size` variable. I have added an assert statement to constrain this value from going above the upper limit.

test/cbmc/proofs/ota_mqtt/OTA_MQTT_strerror/OTA_MQTT_strerror_harness.c
Bujain: I have removed similar comments from the same

test/cbmc/proofs/ota_http/initFileTransfer_Http/initFileTransfer_Http_harness.c
```diff
@@ -0,0 +1,69 @@
...
+
+void initFileTransfer_Http_harness()
+{
+    OtaAgentContext_t * pAgentCtx;
+    OtaFileContext_t fileContext;
+    OtaHttpInterface_t http;
+    OtaHttpStatus_t status;
+    OtaInterfaces_t interface;
+
+    pAgentCtx = ( OtaAgentContext_t * ) malloc( sizeof( OtaAgentContext_t ) );
```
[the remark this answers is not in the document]
Bujain: I have statically allocated a memory address to `pAgentCtx` and removed the use of malloc.

test/cbmc/proofs/ota_http/initFileTransfer_Http/initFileTransfer_Http_harness.c
Bujain: Yes. Static allocation would remove the need to assume that the pAgentCtx can point to NULL.

test/cbmc/proofs/ota_http/decodeFileBlock_Http/decodeFileBlock_Http_harness.c
```diff
@@ -0,0 +1,80 @@
...
+    size_t payloadSize;
+
+    /* Initializing the unconstrained variables. These variables can never be NUll
+     *  since they have been declared statically in the decodeAndStoreDataBlock() function. */
+    pFileId = &fileId;
+    pBlockId = &blockId;
+    pBlockSize = &blockSize;
+
+    /* Allocate memory for the payload and the buffer to store the message. */
+    char * size = ( char * ) malloc( messageSize );
```
yanjos-dev, line 55: Two things.
1. I thought we were removing dynamic memory allocation from the tests?
2. This variable name does not seem to be correct. This "size" variable seems like it's being used to represent the payload. In this case, you could set *pPayload to this malloc result.

I think these can be statically allocated instead of dynamic. The size used in the OTA library is fixed to be ( 1UL << otaconfigLOG2_FILE_BLOCK_SIZE ). It's good to test different values for this though, so it may make sense to set it to this value and then vary the config file in some way. Otherwise you could set it to ( 1UL << variable), where the value of variable is based on the valid range of values for otaconfigLOG2_FILE_BLOCK_SIZE.

test/cbmc/proofs/ota_interface/setDataInterface/setDataInterface_harness.c
```diff
@@ -0,0 +1,56 @@
...
+    OtaErr_t err;
+
+    uint8_t * pProtocol;
+    size_t size;
+
+    /* The maximum size of the protocol buffer is defined by OTA_PROTOCOL_BUFFER_SIZE
+     *  and thus size variable cannot exceed OTA_PROTOCOL_BUFFER_SIZE. */
+    __CPROVER_assume( size <= OTA_PROTOCOL_BUFFER_SIZE );
+
+    /* Initialize the variable. */
```
yanjos-dev, line 43: nitpick: size is not very informative. What is the the size of? Can you update this name?

test/cbmc/proofs/ota_mqtt/OTA_MQTT_strerror/OTA_MQTT_strerror_harness.c
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#include "ota_mqtt_private.h"
+
+void OTA_MQTT_strerror_harness()
+{
+    OtaMqttStatus_t status;
+    const char * pstr;
+
+    pstr = OTA_MQTT_strerror( status );
```
yanjos-dev, line 35: nitpick: This should be "pStr". Str is not very descriptive though, I would recommend something like "pErrMessage".

test/cbmc/proofs/ota_interface/setDataInterface/setDataInterface_harness.c
yanjos-dev: Another nitpick, no need to comment about initializing a variable. I think the user will be able to understand this much just from looking at the code briefly.

PR #244, opened 2021-08-05
Added assert statement to avoid memcpy failure

Bujain:
*Issue #, if available:*
If the size of the payload is less than the messageSize, then the memcpy in the decodeFileBlock function would fail.

*Description of changes
Added a statement to assert the paylaodSize to be greater than messageSize


source/ota.c
```diff
@@ -2586,6 +2586,11 @@ static IngestResult_t decodeAndStoreDataBlock( OtaFileContext_t * pFileContext,
     /* Decode the file block if space is allocated. */
     if( payloadSize > 0u )
     {
+        if( messageSize <= OTA_FILE_BLOCK_SIZE )
+        {
+            assert( ( payloadSize >= messageSize ) );
+        }
```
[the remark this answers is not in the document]
Bujain: Yes. your understanding of the problem is correct. If the messageSize is greater than the OTA_FILE_BLOCK_SIZE; we have a different error handling for this. This can be removed and still check for the problem.

source/ota.c
```diff
@@ -2584,7 +2584,7 @@ static IngestResult_t decodeAndStoreDataBlock( OtaFileContext_t * pFileContext,
     }
 
     /* Decode the file block if space is allocated. */
-    if( payloadSize > 0u )
+    if( ( payloadSize > 0u ) && ( payloadSize >= messageSize ) )
```
yanjos-dev, line 2587: 1. Is it always an error if the payloadSize is not greater than or equal to messageSize? What is the expected behavior here? I'm wondering where messageSize comes from and if it could be smaller in some cases, for example the last block in a download.
2. This check is redundant, if we are requiring payloadSize to be >= message size, then we don't need to also check that it's greater than 0. We can just check that it's greater than or equal to the messageSize (assuming this is what we want to be checking).

yanjos-dev:
The unit test and coverage tests are failing. You will need to update the unit tests to get this to pass if you change the logical flow or branching of the function.

source/ota.c
Bujain: 1. Yes, It's always an error if the messageSize is greater than the payloadSize. 
2. I will change it in the next commit.

Bujain:
This fix is not necessary since the `payloadSize` cannot exceed the `OTA_FILE_BLOCK_SIZE` and their already exists a check between `messageSize` and `OTA_FILE_BLOCK_SIZE`.

PR #246, opened 2021-08-06
Added CBMC proofs for setControlInterface function

Bujain:
*Description of changes:*
Added CBMC proofs for setControlInterface function. The new files added to the repo are:
setControlInterface/Makefile
setControlInterface/setControlInterface_harness.c
setControlInterface/cbmc-viewer.json
setControlInterface/cbmc-proof.txt
setControlInterface/README.md


test/cbmc/proofs/setControlInterface/setControlInterface_harness.c
```diff
@@ -0,0 +1,40 @@
...
+{
+    OtaControlInterface_t * pControlInterface;
+    OtaControlInterface_t controlInterface;
+
+    /* Initialize the control interface. The control interface can never
+     *  be NULL as it is declared statically in ota.c source file. */
+    pControlInterface = &controlInterface;
+
+    setControlInterface( pControlInterface );
+}
```
yanjos-dev, line 40: This test seems like it would only be valiud if the configENABLED_CONTROL_PROTOCOL macro is set to OTA_CONTROL_OVER_MQTT.
https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/813dcbc93e7b1088f890ce1ca023816127f91938/source/ota_interface.c#L67-L71


Can you make it so that the test defines this macro value so we aren't just assuming that the default config file is what it needs to be?

yanjos-dev: This won't work. You defined the macro in the scope of this .c file, so it will not be visible in the ota_interface.c file, where it needs to be used.

You need to define the macro either in a config file file, or a preprocessor definition. I don't recommend making test specific changes to a config file unless you have a way to have separate configs for each test. So you should probably use preprocessor definitions via compiler flags.

Bujain: Updated it in 2e321cf

yanjos-dev (approved):

cobusve (approved):

PR #247, opened 2021-08-06
Added CBMC proofs for OTA_MQTT_strerror function

Bujain:
*Description of changes:*
This PR contains the CBMC proofs for the OTA_MQTT_strerror function. The files added to this PR are:

OTA_MQTT_strerror/Makefile
OTA_MQTT_strerror/cbmc-proof.txt
OTA_MQTT_strerror/cbmc-viewer.json
OTA_MQTT_strerror/OTA_MQTT_strerror_harness.c
OTA_MQTT_strerror/README.md


test/cbmc/proofs/Makefile-template-defines
```diff
@@ -2,7 +2,7 @@
 # Absolute path to the root of the source tree.
 #
 SRCDIR ?= $(abspath $(PROOF_ROOT)/../../../)
-
+>
```
yanjos-dev, line 5: What is this character? Was this added by accident?

test/cbmc/proofs/OTA_MQTT_strerror/OTA_MQTT_strerror_harness.c
```diff
@@ -0,0 +1,40 @@
...
+
+void OTA_MQTT_strerror_harness()
+{
+    OtaMqttStatus_t status;
+    const char * pErrorMsg;
+
+    pErrorMsg = OTA_MQTT_strerror( status );
+
+    /* The function returns a string which is never NULL. If it does, then it is a
+     *  problem. */
```
yanjos-dev, line 38: It's true that based on the implementation, the string can not be returned as a non-NULL value. Kind of a nitpick, but the assertion is more about the expectation of the caller and less about the implementation. The assert is to confirm that the implementation is doing what we expect it to be doing. So I would reword it to be something like this:
```suggestion
    /* The caller of this function assumes that the return value is never NULL. */
```

test/cbmc/proofs/OTA_MQTT_strerror/OTA_MQTT_strerror_harness.c
```diff
@@ -0,0 +1,40 @@
...
+    __CPROVER_assert( pErrorMsg != NULL, "Function return cannot be NULL" );
```
yanjos-dev, line 39: Please make the error string a full sentence that ends in a period. Also, this error string could be more descriptive about what actually happened. "Function return cannot be NULL" is stating an assumption, not an error.

yanjos-dev (changes requested):

test/cbmc/proofs/Makefile-template-defines
Bujain: Removed it in 207cf11

test/cbmc/proofs/OTA_MQTT_strerror/OTA_MQTT_strerror_harness.c
Bujain: Updated it in 207cf11

test/cbmc/proofs/OTA_MQTT_strerror/OTA_MQTT_strerror_harness.c
Bujain: Updated it in 207cf11

yanjos-dev (approved):

aggarg (approved):

PR #253, opened 2021-08-11
Added CBMC proof for stringBuilderUInt32Hex function 

Bujain:
*Description of changes:*

Added the files required for the CBMC-proof for stringBuilderUInt32Hex function. 


test/cbmc/proofs/stringBuilderUInt32Hex/stringBuilderUInt32Hex_harness.c
```diff
@@ -0,0 +1,47 @@
...
+void stringBuilderUInt32Hex_harness()
+{
+    char * pBuffer;
+    size_t bufferSizebytes;
+    uint32_t value;
+
+    /* The bufferSizebytes is the size of the pBuffer. The pBuffer is statically initialized with
+     *  a size of U32_MAX_LEN + 1 in all the functions which call stringBuilderUInt32Decimal. Hence,
+     *  the size can never be below that. */
+    __CPROVER_assume( bufferSizebytes > U32_MAX_LEN );
```
yanjos-dev, line 36: This assumption doesn't quite make sense to me.
1. Do we need an assumption for this?
2. If we do need an assumption, why make it go from the expected size all the way up to the maximum possible value?

I see your comment that it can never be below that size, but does that not also mean that it can't be above that size?

test/cbmc/proofs/stringBuilderUInt32Hex/stringBuilderUInt32Hex_harness.c
```diff
@@ -0,0 +1,47 @@
...
+
+    pBuffer = ( char * ) malloc( bufferSizebytes );
+
+    /* pBuffer is always initialized statically before passing it to the function. Hence,
+     *  it can never be NULL. */
+    __CPROVER_assume( pBuffer != NULL );
```
yanjos-dev, line 42: This doesn't do anything and can be removed since pBuffer is being set right above it.

yanjos-dev (changes requested):

test/cbmc/proofs/stringBuilderUInt32Hex/stringBuilderUInt32Hex_harness.c
Bujain: Yes but sometimes malloc can fail and return a NULL value which is being avoided by using this statement.

test/cbmc/proofs/stringBuilderUInt32Hex/stringBuilderUInt32Hex_harness.c
Bujain: This assumption is required because the size of the pBuffer should always be above the U32_MAX_LEN. If the size of the goes below the size U32_MAX_LEN then it hits the assert statement in the source code. The assumption does not have a max constraint since in any case if the code is modified with the size of pBuffer greater than U32_MAX_LEN the CBMC proofs should not fail.

yanjos-dev: I see, thanks for clarifying.

test/cbmc/proofs/stringBuilderUInt32Hex/stringBuilderUInt32Hex_harness.c
yanjos-dev: We spoke offline, but I'll share for others.

CBMC proofs always generate the permutations of what that variable could be. But if a given permutation doesn't hold the assumptions true, it terminates that branch early. So the behavior of having an assume statement after the malloc is that if malloc returns NULL, then the assume statement will not hold true, and the branch won't be traversed. This equates to the test passing without any failures. I'm okay with this since I don't expect malloc to typically fail for this test.

test/cbmc/proofs/stringBuilderUInt32Hex/stringBuilderUInt32Hex_harness.c
```diff
@@ -0,0 +1,39 @@
...
+    char * pBuffer;
+    size_t bufferSizebytes;
+    uint32_t value;
+
+    /* The bufferSizebytes is the size of the pBuffer. The pBuffer is statically initialized with
+     *  a size of U32_MAX_LEN + 1 in all the functions which call stringBuilderUInt32Decimal. Hence,
+     *  the size can never be below that. */
+    __CPROVER_assume( bufferSizebytes > U32_MAX_LEN );
+
+    pBuffer = ( char * ) malloc( bufferSizebytes );
```
yanjos-dev, line 30:
```suggestion
    pBuffer = ( char * ) malloc( bufferSizebytes * sizeof( char ) );
```

You may also want to change the name of the bufferSizeBytes variable. Since the number of bytes depends on the size of the char type. It should 1 byte in most cases, but it's not enforced by the standard.

yanjos-dev (changes requested):

yanjos-dev (approved):

test/cbmc/proofs/stringBuilderUInt32Hex/stringBuilderUInt32Hex_harness.c
```diff
@@ -0,0 +1,39 @@
...
+
+void stringBuilderUInt32Hex_harness()
+{
+    char * pBuffer;
+    size_t bufferSizebytes;
+    uint32_t value;
+
+    /* The bufferSizebytes is the size of the pBuffer. The pBuffer is statically initialized with
+     *  a size of U32_MAX_LEN + 1 in all the functions which call stringBuilderUInt32Decimal. Hence,
+     *  the size can never be below that. */
```
yanjos-dev, line 27:
```suggestion
    /* The bufferSizebytes is the size of the pBuffer. The pBuffer is statically initialized with
     * a size of U32_MAX_LEN + 1 in all the functions which call stringBuilderUInt32Decimal. Hence,
     * the size can never be below that. */
```

test/cbmc/proofs/stringBuilderUInt32Hex/stringBuilderUInt32Hex_harness.c
```diff
@@ -0,0 +1,39 @@
...
+    __CPROVER_assume( bufferSizebytes > U32_MAX_LEN );
+
+    pBuffer = ( char * ) malloc( bufferSizebytes * sizeof( char ) );
+
+    /* pBuffer is always initialized statically before passing it to the function. Hence,
+     *  it can never be NULL. */
```
yanjos-dev, line 33:
```suggestion
    /* pBuffer is always initialized statically before passing it to the function. Hence,
     * it can never be NULL. */
```

test/cbmc/proofs/stringBuilderUInt32Hex/stringBuilderUInt32Hex_harness.c
```diff
@@ -0,0 +1,39 @@
...
+/* SPDX-License-Identifier: Apache-2.0 */
+
+/**
+ * @file stringBuilderUInt32Hex_harness.c
+ * @brief Implements the proof harness for stringBuilderUInt32Hex function.
+ */
+/* Include files required for mqtt interface. */
+#include "ota_mqtt_private.h"
+
+/*Maximum number of output digits of an unsigned long value.*/
```
yanjos-dev, line 11:
```suggestion
/* Maximum number of output digits of an unsigned long value. */
```

yanjos-dev (changes requested):

yanjos-dev (approved):

aggarg (approved):

PR #254, opened 2021-08-11
Added CBMC proof of stringBuilderUInt32Decimal function

Bujain:
*Description of changes:*
Added files required CBMC proof of stringBuilderUInt32Decimal function


test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
```diff
@@ -0,0 +1,50 @@
+/* Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved. */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+/*
+ * Insert copyright notice
+ */
```
yanjos-dev, line 6: Does this need to be updated to have the copyright notice? If not, delete this.

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * @file stringBuilderUInt32Decimal_harness.c
+ * @brief Implements the proof harness for stringBuilderUInt32Decimal function.
+ */
+
+/*
+ * Insert project header files that
+ *   - include the declaration of the function
+ *   - include the types needed to declare function arguments
+ */
```
yanjos-dev, line 17: Please delete this.

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
```diff
@@ -0,0 +1,50 @@
...
+{
+    /* The pBuffer is used as string in various functions but the maximum size of the. */
+    char * pBuffer;
+    size_t bufferSizebytes;
+    uint32_t value;
+
+    /* The bufferSizebytes is the size of the pBuffer. The pBuffer is statically initialized with
+     *  a size of U32_MAX_LEN + 1 in all the functions which call stringBuilderUInt32Decimal. Hence,
+     *  the size can never be below that. */
+    __CPROVER_assume( bufferSizebytes > U32_MAX_LEN );
```
yanjos-dev, line 39: Is this assumption backwards? This seems like we are assuming that the size is always greater than the max.

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
```diff
@@ -0,0 +1,50 @@
...
+
+    pBuffer = ( char * ) malloc( bufferSizebytes );
+
+    /* pBuffer is always initialized statically before passing it to the function. Hence,
+     *  it can never be NULL. */
+    __CPROVER_assume( pBuffer != NULL );
```
yanjos-dev, line 45: You don't need this assumption. pBuffer is directly being assigned in the line above.

yanjos-dev (changes requested):

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
Bujain: Removed it in 327af27

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
Bujain: Removed it in 327af27

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
Bujain: `U32_MAX_LEN` is not the maximum size of the `pBuffer` but it is the maximum number of digits of an unsigned long value. pBuffer should be atleast of that size to store all the digits in it and also the null character.

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
Bujain: This assumption is required as malloc can fail sometimes and return NULL value. To avoid such error, we need to have this assume statement in the code.

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
yanjos-dev: I see, thank you for clarifying. In this case, do we want to test for the invalid inputs of a buffer size that's smaller? What if the user passes in a buffer of size U32_MAX_LEN?

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
yanjos-dev: Does __CPROVER_assume have any impact on a variable that's been assigned to previously?

My understanding is that the CBMC framework will assign values to variables that have not been assigned. Then if the CBMC framework is assigning a value to the variable, then the __CPROVER_assume function constrains what the assignment can be. Is this not the case?

If it is the case, then pBuffer will always be assigned. Even if it fails, it will be assigned to NULL.

It seems to me like we should have an if check on the result of malloc, or just an assert statement to catch it if it is NULL.

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
Bujain: So the User does not define the buffer. It's defined as a local variable in the static function. So we don't have to worry about that issue

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
Bujain: Yes. The __CPROVER_assume has an impact on a variable that has been assigned to previously. In this case the even though the pBuffer is assigned previously it was not pointing to any memory address and this is taken care by malloc. But then again the malloc can also fail and assign NULL to the pBuffer which is then taken care of by the __CPROVER_assume statement.

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
yanjos-dev: Sounds good, thanks.

test/cbmc/proofs/stringBuilderUInt32Decimal/stringBuilderUInt32Decimal_harness.c
yanjos-dev: Thanks for clarifying, I agree with your approach now.

yanjos-dev (approved):

aggarg (approved):

PR #255, opened 2021-08-12
Added CBMC proof of stringBuilder function

Bujain:
*Description of changes:*
Added all the files for CBMC proof of stringBuilder function,


test/cbmc/proofs/stringBuilder/stringBuilder_harness.c
```diff
@@ -0,0 +1,62 @@
...
+
+    /* Strings array is always passed with a NULL string at the end. */
+    __CPROVER_assume( strings[ numStrings - 1 ] == NULL );
+
+    __CPROVER_file_local_ota_mqtt_c_stringBuilder( pBuffer, bufferSizeBytes, strings );
+
+    /* Free the allocated memory. */
+    free( pBuffer );
+
+    for( i = 0; i < numStrings; ++i )
```
yanjos-dev, line 56: Shouldn't this be `i < ( numStrings - 1)`?

test/cbmc/proofs/stringBuilder/stringBuilder_harness.c
```diff
@@ -0,0 +1,62 @@
...
+
+    /* The pBuffer is statically allocated with a non-zero size before passing it to the stringBuilder function.*/
+    __CPROVER_assume( bufferSizeBytes > 0 );
+    pBuffer = ( char * ) malloc( bufferSizeBytes );
+
+    /* pBuffer can never be NULL since it it always initialized before passing it to the stringBuilder function. */
+    __CPROVER_assume( pBuffer != NULL );
+
+    /* The minimum and the maximum number of strings inside the numStrings is 11. */
+    __CPROVER_assume( numStrings > 0 && numStrings < UINT32_MAX );
```
yanjos-dev, line 34: This wording sounds like you are saying both the min and the max are 11. I think you mean to say that the min is 0 and the max is 11.

Also, if the max is 11, then why have the assumption go to UINT32_MAX? `size_t` is also a signed data type, so it doesn't make sense to compare it to UINT32_MAX.

test/cbmc/proofs/stringBuilder/stringBuilder_harness.c
```diff
@@ -0,0 +1,62 @@
...
+    strings = ( char ** ) malloc( numStrings * sizeof( char * ) );
+
+    /* The strings array cannot be NULL as it is always initialized before
+     *  passing to the function. */
+    __CPROVER_assume( strings != NULL );
+
+    for( i = 0; i < numStrings - 1; ++i )
+    {
+        strings[ i ] = ( char * ) malloc( stringSize );
+    }
```
yanjos-dev, line 46: This condition: `i < numStrings - 1` is allocating only four strings, even if `numStrings `is 5. I think you want `i < numStrings` instead. This would match the previous malloc you did, where you allocated space for `numStrings `(in this ex. 5). This will also match the free statement that you have down below.

test/cbmc/proofs/stringBuilder/stringBuilder_harness.c
```diff
@@ -0,0 +1,62 @@
...
+
+    /* Strings array is always passed with a NULL string at the end. */
+    __CPROVER_assume( strings[ numStrings - 1 ] == NULL );
```
yanjos-dev, line 49: If you index into `strings`, you are getting the pointer to the first byte in the string array. So `strings[ numStrings - 1 ] == NULL` is adding an assumption that the first byte in the last string in the array of strings is not NULL. In other words, that the malloc didn't fail when you allocated the string.

If you want to add an assume statement that each string is NULL terminated, then you could either:
1. When you allocate the string, set the last index in it to 0.
2. Update the assume statement to be something more like `strings[i][stringSize - 1] == NULL` (in a loop).

test/cbmc/proofs/stringBuilder/stringBuilder_harness.c
yanjos-dev, line 56: This should match your previous memory allocation. I think this is correct though and that the malloc code needs to change, as commented above.

yanjos-dev (changes requested):

test/cbmc/proofs/stringBuilder/stringBuilder_harness.c
Bujain: Updated it in 6d32385

test/cbmc/proofs/stringBuilder/stringBuilder_harness.c
Bujain: Yes. I have changed the free statement accordingly. In this proof I only to allocate memory for the first `numStrings-1` pointers.The previous malloc is allocate memory to store the pointers, In strings array the last pointer should always be a NULL pointer and that is why I have assumed the last pointer to be NULL.

test/cbmc/proofs/stringBuilder/stringBuilder_harness.c
Bujain: I did not clearly understand this comment. In the previous commit, I have added the assumption that all the strings in the for loop are allocated properly and the last character in every string is a null character. 

The strings array is an array of character pointers and the statement strings `[ numStrings - 1] == NULL` statement is to add a `NULL` pointer to the last element in the array.

yanjos-dev: After taking another look, I see that "stringBuilder" assumes that the strings input array is null terminated so this makes sense now.

yanjos-dev (approved):

aggarg (approved):

PR #277, opened 2021-08-24
Add CBMC proof of decodeFIleBlock_Mqtt function

Bujain:
Add CBMC proof of decodeFIleBlock_Mqtt function

Description
-----------
Added CBMC proof files of decodeFIleBlock_Mqtt function

Checklist:
----------


test/cbmc/proofs/decodeFileBlock_Mqtt/decodeFileBlock_Mqtt_harness.c
```diff
@@ -0,0 +1,53 @@
...
+    int32_t * pBlockSize;
+    uint8_t ** pPayload;
+    size_t * pPayloadSize;
+
+    decodeFileBlock_Mqtt( pMessageBuffer,
+                          messageSize,
+                          pFileId,
+                          pBlockId,
+                          pBlockSize,
+                          pPayload,
```
yanjos-dev, line 51: I believe this variable needs to be freed after the call to this function. The OTA_CBOR_Decode_GetStreamResponseMessage function called inside mallocs and sets this variable to the result.

test/cbmc/proofs/decodeFileBlock_Mqtt/decodeFileBlock_Mqtt_harness.c
```diff
@@ -0,0 +1,53 @@
...
+        ( pBlockSize == NULL ) ||
+        ( pPayload == NULL ) ||
+        ( pPayloadSize == NULL ) ||
+        ( pMessageBuffer == NULL ) )
+    {
+        status = false;
+    }
+
+    return status;
+}
```
yanjos-dev, line 34: Do we need to mock this function instead of using it directly? If we can use it directly, that would model the behavior of the code better.

In this case this function usually mallocs some memory and then the caller needs to free it. So I think it could make sense to use the original function.

Bujain: IMO it's not required for to call the function directly. Using a stub saves time in reproducing a similar result and is computationally efficient. The called function is also checked for memory safety.

yanjos-dev: Okay if the called function is tested as well then that's okay.

yanjos-dev (approved):

aggarg (approved):

PR #280, opened 2021-08-31
Add CBMC proof of OTA_CBOR_Encode_GetStreamRequestMessage functions

Bujain:
Add CBMC proof of OTA_CBOR_Encode_GetStreamRequestMessage functions

Description
-----------
Added CBMC proof files for OTA_CBOR_Encode_GetStreamRequestMessage functions

Checklist:
----------


.vscode/settings.json
```diff
@@ -1,5 +1,6 @@
 {
     "files.associations": {
-        "ota_mqtt_private.h": "c"
+        "ota_mqtt_private.h": "c",
```
yanjos-dev, line 3: is ota_mqtt_private.h required for this proof?

test/cbmc/proofs/OTA_CBOR_Encode_GetStreamRequestMessage/OTA_CBOR_Encode_GetStreamRequestMessage_harness.c
```diff
@@ -0,0 +1,110 @@
...
+    int32_t fileId;
+    int32_t blockSize;
+    int32_t blockOffset;
+    uint8_t * pBlockBitmap;
+    size_t blockBitmapSize;
+    int32_t numOfBlocksRequested;
+    size_t clientTokenSize;
+
+    /* The pMessageBuffer pointer is pointing to an array of size messageBufferSize. */
+    pMessageBuffer = ( uint8_t * ) malloc( messageBufferSize * sizeof( uint8_t ) );
```
yanjos-dev, line 95: This memory needs to be freed before the function returns. Also, please add an assume statement in case we fail to allocate to memory. Please add this to the rest of the mallocs as well.

yanjos-dev (changes requested):

test/cbmc/proofs/OTA_CBOR_Encode_GetStreamRequestMessage/OTA_CBOR_Encode_GetStreamRequestMessage_harness.c
Bujain: In this proof, we also have to check for cases where the malloc fails. I don't think we require any assume statements in this proof.

.vscode/settings.json
Bujain: Updated in 1b86b2f

test/cbmc/proofs/OTA_CBOR_Encode_GetStreamRequestMessage/OTA_CBOR_Encode_GetStreamRequestMessage_harness.c
yanjos-dev: If we don't check for the return value of malloc, does that mean it will be set to NULL in the tests some of the time?

Bujain: Yes. Sometimes the malloc can fails and when it fails it would return NULL. CBMC would consider this as a test case.

yanjos-dev (approved):

n9wxu (approved):

Issue #217, opened 2021-05-25
event.deinit is never called on shutdown


fhars:
The event.deinit callback is not called on shutdown, and event.init will be called again if you call OTA_Init after the agent has shutdown.

The two examples in os/portable happen to use methods (xQueueCreateStatic and mq_open) where this does not create issues, but in general it may lead to resource leaks, for example if someone happens to compile FreeRTOS without 
 `#define configSUPPORT_STATIC_ALLOCATION  1`
and then replaces that call with xQueueCreate.

pvyawaha:
Hello, Thank you for pointing this out, for non static implementations other than the two provided event.deinit must be called. We are working internally on a pull request to address this.

PR #330, opened 2021-10-25
Deinitialize event in OTA_Shutdown

Bujain:
Deinitialize OS event mechanism in OTA_Shutdown. 

Description
-----------
This PR calls deinit for the event mechanism in the OTA_Shutdown. This is a fix to the issue (#217 ).

Checklist:
----------


source/ota.c
```diff
@@ -2737,6 +2737,7 @@ static IngestResult_t ingestDataBlock( OtaFileContext_t * pFileContext,
 static void agentShutdownCleanup( void )
 {
     uint32_t index;
+    OtaEventContext_t * pEventCtx = NULL;
```
ravibhagavandas, line 2740: Why do we need this local variable on stack, if we are not using it?

ActoryOu:
#440 merged. We don't need to do event.deinit in shutdown stage.

Issue #219, opened 2021-05-25
Race condition and undefined behaviour in `OTA_Shutdown`


fhars:
`OTA_Shutdown` contains this check at the top of the function:
```
    if( otaAgent.state == OtaAgentStateInit )
    {
        /* When in init state, the OTA state machine is not running yet. So directly set state to
         * stopped. */
        otaAgent.state = OtaAgentStateStopped;
    }
```
Since `OTA_Shutdown` is part of the public API, it may be called from any task, so there is no guarantee at all that the `OTA_EventProcessingTask` doesn't start executing somwhere else right after this check, setting `otaAgent.state` to `otaAgentStateReady`. 

Aditionally, since `otaAgent.state` is defined as a plain, non-atomic enum value, there is no guarantee at all that any changes to `otaAgent.state` made by the task executing `OTA_EventProcessingTask` are ever seen by code calling `OTA_Shutdown` from another task. A sufficiently smart compiler that understood FreeRTOS tasks and saw e.g. a device state management task that only ever called `OTA_Init`, `OTA_Suspend`, `OTA_Resume` and `OTA_Shutdown`  would be perfectly free to conclude that the only values `otaAgent.state` could ever have are `OtaAgentStateStopped` and `OtaAgentStateInit` and that the logic to actually send the shutdown event is dead code and inline an optimized version (which in that scenario without log messages would just be `otaAgent.state = OtaAgentStateStopped; return otaAgent.state;`) for that task.

divekarshubham:
Thank you for bringing the issue to our attention, we are working on a timeline to incorporate this change into the library.

PR #331, opened 2021-10-26
Add shutdown flag for Ota_EventProcessing

Bujain:
Add shutdown flag for Ota_EventProcessing

Description
-----------
This PR adds a shutdown flag to prevent the race condition between Ota_EventProcessingTask and Ota_Shutdown. It also adds a volatile qualifier to the state field in OtaAgentContext_t. These are the fixes for #219 

Checklist:
----------


ActoryOu:
We don't need this shutdown flag because we leave a event in the event queue to handle the shutdown event even it's in init stage to avoid the scenario below.
1. One another task "A" is ready to run Ota_EventProcessingTask
2. Shutdown flow is on-going on task "B", reset the function pointer but the state is not set to shutdown yet.
3. "A" call the os.timer.start in some handler, but the timer should be stopped because it's going to shutdown.

PR #362, opened 2021-11-15
Add CBMC proof of handleSelfTestJobDoc

Bujain:
Add CBMC proof of handleSelfTestJobDoc function

Description
-----------
This PR adds files required for the CBMC proof of handleSelfTestJobDoc function. 

Checklist:
----------

- [ ] I have read and applied the rules stated in CONTRIBUTING.md. 


test/cbmc/source/stubs.c
```diff
@@ -0,0 +1,152 @@
...
+
+    return err;
+}
+
+OtaOsStatus_t startTimerStub( OtaTimerId_t otaTimerId,
+                              const char * const pTimerName,
+                              const uint32_t timeout,
+                              OtaTimerCallback_t callback )
+{
+    OtaOsStatus_t status;
```
gshvang, line 48: Status should be initialized.

test/cbmc/source/stubs.c
```diff
@@ -0,0 +1,152 @@
...
+    __CPROVER_assert( pTimerName != NULL, "Error: TimerName cannot be NULL" );
+
+    __CPROVER_assert( callback != NULL, "Error: callback should be pointing to a function." );
+
+    return status;
+}
+
+OtaOsStatus_t stopTimerStub( OtaTimerId_t otaTimerId )
+{
+    OtaOsStatus_t status;
```
gshvang, line 62: Status should be initialized.

test/cbmc/source/stubs.c
```diff
@@ -0,0 +1,152 @@
...
+
+    __CPROVER_assert( pFileContext != NULL,
+                      "Error: pFileContext in the otaAgent is statically initialized and hence cannot be NULL." );
+
+    return state;
+}
+
+void otaAppCallback( OtaJobEvent_t eEvent,
+                     const void * pData )
+{
```
gshvang, line 85: Return statement missing here

gshvang (approved):

gshvang (approved):

test/cbmc/proofs/handleSelfTestJobDoc/handleSelfTestJobDoc_harness.c
```diff
@@ -0,0 +1,57 @@
...
+ */
+/*  Ota Agent includes. */
+#include "ota.h"
+#include "stubs.h"
+
+extern OtaAgentContext_t otaAgent;
+extern void handleSelfTestJobDoc( const OtaFileContext_t * pFileContext );
+
+OtaErr_t validateUpdateVersion( const OtaFileContext_t * pFileContext )
+{
```
paulbartell, line 35: I don't think this override is necessary. 
handleSelfTestJobDoc() should validate it's parameter via an assertion.

Bujain: @paulbartell In this case an assertion to validate the parameter is not required since the test function is an internal function and will be called with a statically allocated variable. The override was required because if there is no body for validateUpdateVersion then CBMC would assume `OtaErr_t` as an `unsigned int` and return values which are out of the values from the enum.

paulbartell: So could a solar flare. That's kinda the point.

test/cbmc/proofs/handleSelfTestJobDoc/handleSelfTestJobDoc_harness.c
```diff
@@ -0,0 +1,57 @@
...
+    return err;
+}
+
+void handleSelfTestJobDoc_harness()
+{
+    OtaFileContext_t fileContext;
+    OtaInterfaces_t otaInterface;
+
+    otaInterface.pal.reset = resetPalStub;
+    otaAgent.OtaAppCallback = otaAppCallbackStub;
```
paulbartell, line 50: The rest of the otaAgent struct should probably be unconstrained right?

test/cbmc/proofs/handleSelfTestJobDoc/handleSelfTestJobDoc_harness.c
AniruddhaKanhere: I am fine either way - have a stub or don't have a stub. As long as this function is checked in another proof, I am happy.
It is up to you @paulbartell :)

test/cbmc/proofs/handleSelfTestJobDoc/handleSelfTestJobDoc_harness.c
```diff
@@ -0,0 +1,57 @@
...
+ * @brief Implements the proof harness for handleSelfTestJobDoc function.
+ */
+/*  Ota Agent includes. */
+#include "ota.h"
+#include "stubs.h"
+
+extern OtaAgentContext_t otaAgent;
+extern void handleSelfTestJobDoc( const OtaFileContext_t * pFileContext );
+
+OtaErr_t validateUpdateVersion( const OtaFileContext_t * pFileContext )
```
AniruddhaKanhere, line 34: I would add an assertion on this pointer being non-NULL.

test/cbmc/proofs/handleSelfTestJobDoc/handleSelfTestJobDoc_harness.c
paulbartell: Is there a particular reason to constrain the return value of validateUpdateVersion? It looks like the current implementation of handleSelfTestJobDoc could handle any value.

test/cbmc/proofs/handleSelfTestJobDoc/handleSelfTestJobDoc_harness.c
```diff
@@ -0,0 +1,57 @@
...
+
+    /* err can assume values of OtaErr_t enum. */
+    __CPROVER_assume( ( err >= OtaErrNone ) && ( err <= OtaErrActivateFailed ) );
+
+    return err;
+}
+
+void handleSelfTestJobDoc_harness()
+{
+    OtaFileContext_t fileContext;
```
AniruddhaKanhere, line 46: Also, it would be good if havoc the data in the `otaAgent` object like so:
`__CPROVER_havoc_object( &otaAgent )`.

Since it is initialized, CBMC will not override it in my knowledge. Thus, havoc-ing it would give us more coverage.

AniruddhaKanhere (changes requested):

test/cbmc/proofs/handleSelfTestJobDoc/handleSelfTestJobDoc_harness.c
Bujain: Since all the of the fields from the otaAgent struct are not being used in the execution of the test function, It's not required to have all the fields of the otaAgent to be unconstrained.

test/cbmc/proofs/handleSelfTestJobDoc/handleSelfTestJobDoc_harness.c
AniruddhaKanhere: No particular reason. If we havoc the `otaAgent`, which we should, then we can call the function in the library. And since the call-tree is not huge, we can get away with this (if it were huge, we'd need to prune some branches by stubbing them out or else the proofs would take too much time to complete). Otherwise, we'd need a stub to return a range of values.

test/cbmc/proofs/handleSelfTestJobDoc/handleSelfTestJobDoc_harness.c
Bujain: ` __CPROVER_havoc_object() ` is used here to non-deterministically set all the bytes in the otaAgent structure.

AniruddhaKanhere (dismissed):
Thanks for making the changes @bujain. I approve the PR.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "files.associations": {
+        "stubs.h": "c"
+    }
+}
```
paulbartell, line 5: File should have an empty line at the end.

test/cbmc/proofs/handleSelfTestJobDoc/handleSelfTestJobDoc_harness.c
paulbartell: @Bujain : Yes. This is not particularly efficient and points out the downside of using globals in this way.

AniruddhaKanhere (approved):

paulbartell (approved):

Issue #412, opened 2022-01-12
OTA_Shutdown does not clear timers (then leading to null pointer dereference)


lightblu:
In contrast to the docs for `OTA_Shutdown` stating "and clear all resources." that does not seem to care for any running timers.

This has been observed and reproduced when the MQTT operation for requesting a new job fails after OTA agent startup (e.g. due to disconnection or a shutdown of the connection meanwhile) which makes the OTA lib start the `OtaRequestTimer`. 
And if then `OTA_Shutdown` is called, that leaves the timer running beyond OTA agent having shut down.

This is especially bad as `OTA_Shutdown` clears the whole `otaAgent`struct, including the `pOtaInterface`pointer.
So when the timer fires, it goes via `otaTimerCallback` into `OTA_SignalEvent` which then does a `otaAgent.pOtaInterface->os.event.send( NULL, pEventMsg, 0 );`
(Note: The potential for `OTA_SignalEvent` has already been pointed out here [OTA_SignalEvent has undefined behaviour](https://github.com/aws/ota-for-aws-iot-embedded-sdk/issues/218) )

Besides this bad crash (kind of sad and funny: the os.event.send function is the second function in the whole struct, thus indexes to a 0x4 from `pOtaInterface`, so if `pOtaInterface` is `NULL`, this actually calls the after-reset-entry-point on our Kinetis ARM M4s, so this didn't even look like the usual rare HardFault but disguised as a very strange rare unexpected reset :( ), this will memory leak created timers.

Note, `OTA_Suspend` does a `otaAgent.pOtaInterface->os.timer.stop( OtaRequestTimer );`, but not `OTA_Shutdown`. However  glancing over the code, it seems this scenario would also be true for the second `OtaSelfTestTimer` which is not even cleared by `OTA_Suspend`, if this is revisited, the whole timer logic needs to be checked.

(Further, I'd appreciate if `ota_os_freertos.c` would make use of static timers not only to avoid dynamic memory leaks on another level, but also because this is one of the last things that would allow the mqtt version to run without heap).

_(This has been observed and confirmed on a 3.1.0 from July - I see though nothing in latest develop til now that touched this)._

archigup:
Hi, we are looking into the issue; I have made a PR to add clearing timers.

Would you be able to make a separate issue for using static timers?

archigup:
[...]

PR #413, opened 2022-01-12
Clear timers in shutdown and suspend

archigup:
Clears OtaRequestTimer and OtaSelfTestTimer in OTA_Shutdown and OTA_Suspend to address null pointer dereferences in the timers if they occur after pOtaInterface is cleared.

Addresses https://github.com/aws/ota-for-aws-iot-embedded-sdk/issues/412


source/ota.c
```diff
@@ -3186,6 +3186,12 @@ OtaState_t OTA_Shutdown( uint32_t ticksToWait,
     {
         otaAgent.unsubscribeOnShutdown = unsubscribeFlag;
 
+        /* Stop the request timer. */
```
divekarshubham, line 3189: NIT: consider moving to the end of this block after actual shutdown.

source/ota.c
```diff
@@ -3386,6 +3392,9 @@ OtaErr_t OTA_Suspend( void )
         /* Stop the request timer. */
         ( void ) otaAgent.pOtaInterface->os.timer.stop( OtaRequestTimer );
 
+        /* Stop the self-test timer. */
```
divekarshubham, line 3395: If we are stopping the timer here, we need to restart it in `OTA_Resume()`

source/ota.c
archigup: This needs to occur before shutdown is queued so that it will always happen before the struct is cleared,

source/ota.c
```diff
@@ -3383,8 +3389,13 @@ OtaErr_t OTA_Suspend( void )
     /* Check if OTA Agent is running. */
     if( otaAgent.state != OtaAgentStateStopped )
     {
-        /* Stop the request timer. */
+        /* Stop and delete the request timer. */
         ( void ) otaAgent.pOtaInterface->os.timer.stop( OtaRequestTimer );
+        ( void ) otaAgent.pOtaInterface->os.timer.delete( OtaRequestTimer );
```
divekarshubham, line 3394: Need to delete the timer while shutdown. Not required for suspend.

divekarshubham (approved):

divekarshubham (approved):

aggarg (approved):

PR #430, opened 2022-05-17
Fix job ID length before memcpy. Call OtaAppCallback only when job I…

ActoryOu:
- Job ID sould reserve a space for NULL. Fix the condition in custom job.
- Call OtaAppCallback only when job ID is valid, or we can't send the job update message to the cloud.

Description
-----------

- Fix the length judgement in custom job.
- Move OtaAppCallback to make sure job ID is valid.

Checklist:
----------


n9wxu (approved):

test/unit-test/ota_utest.c
```diff
@@ -3292,3 +3296,37 @@ void test_OTA_packetsProcessedOverflow()
...
+ * job name downloaded using OTA is greater than the maximum size supported by the library. */
+void test_OTA_jobIdMaxLength()
+{
+    pOtaJobDoc = JOB_DOC_INVALID_JOB_ID_LEN_MAX;
+
+    otaGoToState( OtaAgentStateWaitingForJob );
+    TEST_ASSERT_EQUAL( OtaAgentStateWaitingForJob, OTA_GetState() );
+
+    otaReceiveJobDocument();
+    receiveAndProcessOtaEvent();
```
ravibhagavandas, line 3310: For future: It would be good if the function can return an error so that just the indented functionality can be tested here.

ravibhagavandas (approved):

Issue #238, opened 2021-08-02
Change the default signature size to support RSA3072 or RSA4096


NickMcKendree:
As of May 2021, Digicert has stopped issuing 2048 bit Code Signing Certificates and has moved to 3072 bit Certificates. Their notification page indicates this is an industry wide standards change so the AWS OTA library should follow the change. 

The default setting of the OTA library is 256 bytes, which only supports RSA-2048. I pulled the latest main to verify that this is still the default setting. Unless it is a breaking issue somewhere else, should kOTA_MaxSignatureSize be defined for RSA4096 to remain flexible for future changes (who knows how long RSA3072 will remain the standard). 

The change is simple...
In ota_private.h, 
`#define kOTA_MaxSignatureSize           512/* Max bytes supported for a file signature (4096 bit RSA is 512 bytes). */`
or
`#define kOTA_MaxSignatureSize           384/* Max bytes supported for a file signature (3072 bit RSA is 384 bytes). */`

https://knowledge.digicert.com/alerts/code-signing-new-minimum-rsa-keysize.html

pvyawaha:
Hello NickMcKendree,

Thank you for bringing this to notice, we are working on a change for this.

ActoryOu:
The default signature size is changed to 384 to support RSA-3072 after #432 merged.
I'm going to close this issue. Thanks for your reminding.

PR #432, opened 2022-05-19
change default signature size to 384 tto support RSA-3072

ActoryOu:
change default signature size to 384 to support RSA-3072

Description
-----------

As mentioned in #238, change default signature size to 384 to support RSA-3072.

Checklist:
----------

### DIFF
--- a/source/include/ota_private.h
+++ b/source/include/ota_private.h
@@ -369,7 +369,7 @@ typedef struct
 {
     uint16_t size;                         /*!< @brief Size, in bytes, of the signature. */
     uint8_t data[ kOTA_MaxSignatureSize ]; /*!< @brief The binary signature data. */
-} Sig256_t;
+} Sig_t;
 
 /**
  * @ingroup ota_struct_types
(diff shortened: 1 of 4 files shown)


source/include/ota_private.h
```diff
@@ -119,7 +119,7 @@
  * @brief A composite cryptographic signature structure able to hold our largest supported signature.
  */
 
-#define kOTA_MaxSignatureSize           256 /* Max bytes supported for a file signature (2048 bit RSA is 256 bytes). */
```
ravibhagavandas, line 122: I think we should also change the struct name to match the change:

https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/1f2362f005f39a1847d645660ecaff93890cfcd7/source/include/ota_private.h#L372

ravibhagavandas, line 122: I think we should also change the struct name to match the change:

https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/1f2362f005f39a1847d645660ecaff93890cfcd7/source/include/ota_private.h#L372

ActoryOu: Fix in new commit.

ActoryOu: Fixed in new commit.

ravibhagavandas (approved):

PR #433, opened 2022-05-19
Replace mq_send/mq_receive with mq_timedsend/mq_timedreceive

ActoryOu:
Replace mq_send/mq_receive with mq_timedsend/mq_timedreceive.

Description
-----------

To support timeout in posix, replace mq_send/mq_receive with mq_timedsend/mq_timedreceive.
Tested by merging changes to aws-iot-device-sdk-embedded-C, pass OTA job in ota_demo_core_mqtt.

Checklist:
----------

### DIFF
--- a/source/portable/os/ota_os_posix.c
+++ b/source/portable/os/ota_os_posix.c
@@ -53,6 +53,11 @@
 #define MAX_MESSAGES      10
 #define MAX_MSG_SIZE      sizeof( OtaEventMsg_t )
 
+/* Time conversion */
+#define OTA_TIME_MS_TO_S( ms )           ( ms / 1000 )
+#define OTA_TIME_MS_TO_NS( ms )          ( ms * 1000000 )
+#define OTA_TIME_MS_LESS_THAN_S( ms )    ( ms % 1000 )
+
 static void requestTimerCallback( union sigval arg );
 static void selfTestTimerCallback( union sigval arg );
 
@@ -117,6 +122,8 @@ OtaOsStatus_t Posix_OtaSendEvent( OtaEventContext_t * pEventCtx,
                                   unsigned int timeout )
 {
     OtaOsStatus_t otaOsStatus = OtaOsSuccess;
+    struct timespec ts = { 0 };
+    time_t currentTime = time( NULL );
 
     ( void ) pEventCtx;
     ( void ) timeout;
@@ -124,7 +131,17 @@ OtaOsStatus_t Posix_OtaSendEvent( OtaEventContext_t * pEventCtx,
     /* Send the event to OTA event queue.*/
     errno = 0;
 
-    if( mq_send( otaEventQueue, pEventMsg, MAX_MSG_SIZE, 0 ) == -1 )
+    /* Set send timeout. */
+    ts.tv_nsec = OTA_TIME_MS_TO_NS( OTA_TIME_MS_LESS_THAN_S( timeout ) );
+    ts.tv_sec = OTA_TIME_MS_TO_S( timeout );
+
+    /* Detect overflow. */
+    if( ( currentTime > 0 ) && ( ( int64_t ) ( INT32_MAX - ts.tv_sec ) >= ( int64_t ) currentTime ) )
+    {
+        ts.tv_sec += currentTime;
+    }
+
+    if( mq_timedsend( otaEventQueue, pEventMsg, MAX_MSG_SIZE, 0, &ts ) == -1 )
     {
         otaOsStatus = OtaOsEventQueueSendFailed;
 
@@ -150,6 +167,8 @@ OtaOsStatus_t Posix_OtaReceiveEvent( OtaEventContext_t * pEventCtx,
     OtaOsStatus_t otaOsStatus = OtaOsSuccess;
     char * pDst = pEventMsg;
     char buff[ MAX_MSG_SIZE ];
+    struct timespec ts = { 0 };
+    time_t currentTime = time( NULL );
 
     ( void ) pEventCtx;
     ( void ) timeout;
@@ -157,16 +176,33 @@ OtaOsStatus_t Posix_OtaReceiveEvent( OtaEventContext_t * pEventCtx,
     /* Receive the next event from OTA event queue.*/
     errno = 0;
 
-    if( mq_receive( otaEventQueue, buff, sizeof( buff ), NULL ) == -1 )
+    /* Set receive timeout. */
+    ts.tv_nsec = OTA_TIME_MS_TO_NS( OTA_TIME_MS_LESS_THAN_S( timeout ) );
+    ts.tv_sec = OTA_TIME_MS_TO_S( timeout );
+
+    /* Detect overflow. */
+    if( ( currentTime > 0 ) && ( ( int64_t ) ( INT32_MAX - ts.tv_sec ) >= ( int64_t ) currentTime ) )
+    {
+        ts.tv_sec += currentTime;
+    }
+
+    if( mq_timedreceive( otaEventQueue, buff, sizeof( buff ), NULL, &ts ) == -1 )
     {
         otaOsStatus = OtaOsEventQueueReceiveFailed;
 
-        LogError( ( "Failed to receive OTA Event: "
-                    "mq_reqeive returned error: "
-                    "OtaOsStatus_t=%i "
-                    ",errno=%s",
-                    otaOsStatus,
-                    strerror( errno ) ) );
+        if( errno == ETIMEDOUT )
+        {
+            LogDebug( ( "Failed to receive OTA Event before timeout" ) );
+        }
+        else
+        {
+            LogError( ( "Failed to receive OTA Event: "
+                        "mq_timedreceive returned error: "
+                        "OtaOsStatus_t=%i "
+                        ",errno=%s, errno=%d",
+                        otaOsStatus,
+                        strerror( errno ), errno ) );
+        }
     }
     else
     {
--- a/source/include/ota_os_interface.h
+++ b/source/include/ota_os_interface.h
@@ -148,7 +148,7 @@ typedef OtaOsStatus_t ( * OtaSendEvent_t )( OtaEventContext_t * pEventCtx,
  *
  * @param[pEventMsg]     Pointer to store message.
  *
- * @param[timeout]       The maximum amount of time the task should block.
+ * @param[timeout]       The maximum amount of time (msec) the task should block.
  *
  * @return               OtaOsStatus_t, OtaOsSuccess if success , other error code on failure.
  */
--- a/test/unit-test/ota_os_posix_utest.c
+++ b/test/unit-test/ota_os_posix_utest.c
@@ -194,3 +194,29 @@ void test_OTA_posix_MemoryAllocAndFree( void )
 
     STDC_Free( buffer );
 }
+
+/**
+ * @brief Test that the event queue receive timeout.
+ */
+void test_OTA_posix_RecvEventTimeout( void )
+{
+    OtaEventMsg_t otaEventToRecv = { 0 };
+    OtaErr_t result = OtaErrUninitialized;
+    time_t recvTimeoutMs = 3000;
+    time_t timeBeforeTestSec = 0;
+    time_t timeAfterTestSec = 0;
+
+    result = event.init( event.pEventContext );
+    TEST_ASSERT_EQUAL( OtaErrNone, result );
+
+    timeBeforeTestSec = time( NULL );
+    result = event.recv( event.pEventContext, &otaEventToRecv, recvTimeoutMs );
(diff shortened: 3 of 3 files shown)


source/portable/os/ota_os_posix.c
```diff
@@ -150,23 +167,42 @@ OtaOsStatus_t Posix_OtaReceiveEvent( OtaEventContext_t * pEventCtx,
...
-    if( mq_receive( otaEventQueue, buff, sizeof( buff ), NULL ) == -1 )
+    /* Set receive timeout. */
+    ts.tv_nsec = OTA_TIME_MS_TO_NS( OTA_TIME_MS_LESS_THAN_S( timeout ) );
+    ts.tv_sec = OTA_TIME_MS_TO_S( timeout );
+
+    /* Detect overflow. */
+    if( ( currentTime > 0 ) && ( ( int64_t ) ( INT32_MAX - ts.tv_sec ) >= ( int64_t ) currentTime ) )
+    {
+        ts.tv_sec += currentTime;
+    }
```
ravibhagavandas, line 187: if it overflows we should cap to max value?

ActoryOu: Fixed in new commit.

ravibhagavandas (dismissed):

source/portable/os/ota_os_posix.c
```diff
@@ -53,6 +53,11 @@
 #define MAX_MESSAGES      10
 #define MAX_MSG_SIZE      sizeof( OtaEventMsg_t )
 
+/* Time conversion */
+#define OTA_TIME_MS_TO_S( ms )           ( ms / 1000 )
```
paulbartell, line 57: These macros are local to this file, so prefixing with OTA_TIME isn't necessary.

source/portable/os/ota_os_posix.c
```diff
@@ -117,14 +122,30 @@ OtaOsStatus_t Posix_OtaSendEvent( OtaEventContext_t * pEventCtx,
                                   unsigned int timeout )
 {
     OtaOsStatus_t otaOsStatus = OtaOsSuccess;
+    struct timespec ts = { 0 };
+    time_t currentTime = time( NULL );
```
paulbartell, line 126: MISRA Rule 21.10 does not allow the use of time.h functions because they are easily misunderstood. This use case is one such very frequent mistake. On linux, there isn't really an alternative to time.h, so we should probably add a deviation to this rule if one does not already exist.

The time returned by calls to time() is meant for absolute timing rather than relative timing. 
This is because it may be adjust backwards or forwards at any time at a rate that does not match realtime. For relative timing purposes, clock_gettime with CLOCK_MONOTONIC is the best choice.

Regardless, mq_timedsend takes an absolute timeout, so a call to time() or clock_gettime() is not necessary. Please review the man page.

ActoryOu: Thanks for suggestion. I will add the deviation for using time.h.

I have two questions about mq_timedsend/mq_timedreceive. 
1. These functions need the absolute time comparing to the current time. How can I get the absolute timeout without calling time() or clock_gettime()?
2. I tried to replace time() with clock_gettime( CLOCK_MONOTONIC, ... ), but I found that mq_timedreceive returns immediately. I can't use CLOCK_MONOTONIC as the input to mq_timedreceive because it uses CLOCK_REALTIME as timer. The return value of CLOCK_REALTIME is larger than CLOCK_MONOTONIC, so it returns immediately. That way, is it still better to use clock_gettime( CLOCK_REALTIME, .. ) than time( NULL )?

Thanks.

source/portable/os/ota_os_posix.c
ActoryOu: Fixed.

source/portable/os/ota_os_posix.c
```diff
@@ -150,23 +175,46 @@ OtaOsStatus_t Posix_OtaReceiveEvent( OtaEventContext_t * pEventCtx,
...
+    ts.tv_sec = MS_TO_S( timeout );
+
+    /* Detect overflow. */
+    if( ( currentTime > 0 ) && ( ( int64_t ) ( INT32_MAX - ts.tv_sec ) >= ( int64_t ) currentTime ) )
+    {
+        ts.tv_sec += currentTime;
+    }
+    else
+    {
+        ts.tv_sec = INT32_MAX;
```
paulbartell, line 198: On most recent systems, time_t is 64 bits. On linux this is defined by the [TIME_T_MAX macro](https://elixir.bootlin.com/linux/v3.3/source/include/linux/time.h#L43). 

On linux, timespec_add_safe is available which could be a helpful reference.

source/portable/os/ota_os_posix.c
paulbartell: This is suboptimal though. Embedded systems often lack an RTC due to cost or battery safety concerns and can have wildly varying CLOCK_REALTIME values after boot-up. 

Commonly, systems sync time after bootup with an NTP server. However, if the mq_timedsend operation were to occur at exactly the right time, it could return early. Blocking for longer than expected is not very likely but could still occur if ntp adjusts CLOCK_REALTIME backwards significantly.

We should consider replacing posix message queues with unix domain sockets (AF_UNIX, SOCK_DGRAM) since those can be used with a monotonic timebase and are still quite widely supported on unix-like operating systems.

ActoryOu: Mqueue supports epoll/poll and select in Linux. So I use poll to detect if send/recv is available to do with timeout 0 in the latest commit.

source/portable/os/ota_os_posix.c
ActoryOu: No need to calculate absolute time because poll uses relarive time.

source/portable/os/ota_os_posix.c
```diff
@@ -28,6 +28,10 @@
 /* Standard Includes.*/
 #include <stdlib.h>
 #include <string.h>
+
+/* MISRA rule 21.10 prohibits the use of time.h because it is implementation dependent or unspecified.
+ * However, this implementation is on POSIX, which has well defined behavior. */
+/* coverity[misra_c_2012_rule_21_10_violation] */
```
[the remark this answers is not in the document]
ActoryOu: In the latest solution, I don't use clock_gettime to get any absolute time.

source/portable/os/ota_os_posix.c
```diff
@@ -150,30 +177,75 @@ OtaOsStatus_t Posix_OtaReceiveEvent( OtaEventContext_t * pEventCtx,
...
+    }
+
+    /* Receive the message from queue. */
+    if( otaOsStatus == OtaOsSuccess )
+    {
+        if( mq_timedreceive( otaEventQueue, buff, sizeof( buff ), NULL, &ts ) == -1 )
+        {
+            otaOsStatus = OtaOsEventQueueReceiveFailed;
+
+            LogError( ( "Failed to receive OTA Event: "
```
ravibhagavandas, line 237: Probably should not Log Error as it can cause noisy logs when OTA Agent is idle.

source/portable/os/ota_os_posix.c
```diff
@@ -150,30 +177,75 @@ OtaOsStatus_t Posix_OtaReceiveEvent( OtaEventContext_t * pEventCtx,
...
+        errno = 0;
+
+        fds.fd = otaEventQueue;
+        fds.events = POLLIN;
+
+        pollResult = poll( &fds, 1, ( int ) timeout );
+
+        if( pollResult == 0 )
+        {
+            otaOsStatus = OtaOsEventQueueReceiveFailed;
```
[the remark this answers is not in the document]
ActoryOu: Summary from the [manual page](https://man7.org/linux/man-pages/man2/poll.2.html)
I think "0" is the error code for timeout in poll.

>    On success, poll() returns a nonnegative value which is the
       number of elements in the pollfds whose revents fields have been
       set to a nonzero value (indicating an event or an error).  **_A
       return value of zero indicates that the system call timed out
       before any file descriptors became read._**
       
There is no clear description for OtaReceiveEvent_t how to return if timeout. From the current code, we should return fail if timeout or OTA continues the next step with invalid event message.

In receiveAndProcessOtaEvent,
``` 
if( otaAgent.pOtaInterface->os.event.recv( NULL, &eventMsg, configOTA_POLLING_EVENTS_TIMEOUT_MS ) == OtaOsSuccess )
{
    //Do event handler.
}
```

ravibhagavandas (dismissed):

n9wxu (dismissed):

source/portable/os/ota_os_posix.c
```diff
@@ -117,27 +122,49 @@ OtaOsStatus_t Posix_OtaSendEvent( OtaEventContext_t * pEventCtx,
                                   unsigned int timeout )
```
paulbartell, line 122: mq_open should be called with the O_NONBLOCK flag.

source/portable/os/ota_os_posix.c
```diff
@@ -150,30 +177,75 @@ OtaOsStatus_t Posix_OtaReceiveEvent( OtaEventContext_t * pEventCtx,
...
+        else
+        {
+            /* Do nothing in success case. */
+        }
+    }
+
+    /* Receive the message from queue. */
+    if( otaOsStatus == OtaOsSuccess )
+    {
+        if( mq_timedreceive( otaEventQueue, buff, sizeof( buff ), NULL, &ts ) == -1 )
```
paulbartell, line 233: If opened with O_NONBLOCK, mq_timedreceive could instead be mq_receive without the timespec and such.

source/portable/os/ota_os_posix.c
```diff
@@ -117,27 +122,49 @@ OtaOsStatus_t Posix_OtaSendEvent( OtaEventContext_t * pEventCtx,
...
     {
         otaOsStatus = OtaOsEventQueueSendFailed;
 
-        LogError( ( "Failed to send event to OTA Event Queue: "
-                    "mq_send returned error: "
-                    "OtaOsStatus_t=%i "
-                    ",errno=%s",
-                    otaOsStatus,
-                    strerror( errno ) ) );
+        LogError( ( "Invalid input, Posix_OtaSendEvent supports timeout < INT32_MAX, "
```
paulbartell, line 135:
```suggestion
        LogError( ( "Invalid input, Posix_OtaSendEvent supports timeout < INT_MAX, "
```

source/portable/os/ota_os_posix.c
paulbartell, line 122: ```unsigned int timeout``` makes it not possible to block forever. Is that intentional?

source/portable/os/ota_os_posix.c
paulbartell: mq_receive should also be in the "else" success case above.

source/portable/os/ota_os_posix.c
```diff
@@ -117,27 +122,49 @@ OtaOsStatus_t Posix_OtaSendEvent( OtaEventContext_t * pEventCtx,
...
     else
     {
-        LogDebug( ( "OTA Event Sent." ) );
+        /* Send the event to OTA event queue.*/
+        errno = 0;
+
+        fds.fd = otaEventQueue;
+        fds.events = POLLOUT;
+
+        pollResult = poll( &fds, 1, ( int ) timeout );
```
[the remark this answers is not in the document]
paulbartell: I missed line 145. This looks fine.

paulbartell (changes requested):
mq_timedsend / mq_timedreceive should be replaced with mq_send / mq_receive when using non-blocking io.

source/portable/os/ota_os_posix.c
paulbartell, line 147: The poll / mq_send calls should be in a timeout loop in a multiple-producer, single-consumer situation, to account for a race condition when two threads are both blocked sending to the same queue.

source/portable/os/ota_os_posix.c
ActoryOu: That's the prototype of OtaSendEvent_t.
`typedef OtaOsStatus_t ( * OtaSendEvent_t )( OtaEventContext_t * pEventCtx,
                                            const void * pEventMsg,
                                            unsigned int timeout );`

ActoryOu: Fixed.

source/portable/os/ota_os_posix.c
ActoryOu: Fixed.

source/portable/os/ota_os_posix.c
ActoryOu: Fixed.

source/portable/os/ota_os_posix.c
ActoryOu: Thanks for suggestion! Added loop to handle race conditions.
BTW, add pollAndSend and pollAndReceive internal functions to reduce code complexity level.

ravibhagavandas (approved):

PR #436, opened 2022-05-23
Update OtaPalWriteBlock_t comment with a note about returning from the function and exception condition.

ActoryOu:
Update OtaPalWriteBlock_t comment with a note about returning from the function and exception condition.

Description
-----------

Checklist:
----------


pvyawaha (dismissed):

source/ota.c
```diff
@@ -2567,6 +2572,39 @@ static IngestResult_t processDataBlock( OtaFileContext_t * pFileContext,
     return eIngestResult;
 }
 
+/* Store the data block in file context. */
+static IngestResult_t writeDataBlockToFile( OtaFileContext_t * pFileContext,
```
paulbartell, line 2576: Is a separate function really necessary here?

ActoryOu: If I add an assertion in original processDataBlock function, the complexity level is 9.
So I add this separate function to reduce the level.

ActoryOu: If I add an assertion in original processDataBlock function, the complexity level is 9.
So I add this separate function to reduce the level.

paulbartell: I pushed some changes to address this. Please look them over.

ActoryOu: Thanks. That's great!
I also added an unit test case for the coverage.

paulbartell (approved):

pvyawaha (approved):

Issue #218, opened 2021-05-25
OTA_SignalEvent has undefiend behaviour


fhars:
This line:
https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/0d6f669d4348a7130811c4d7c260218aea6534c2/source/ota.c#L2933
may dereference a null pointer, especially after an `OTA_Shutdown(ticks, false)`. (Note that checking that pOtaInterface is valid before the call is not sufficient here, as the `shutdownHandler` might memset `otaAgent` to zero between the check and the call). Another possible race condition is in the called `send` method, as (if the FreeRTOS case) the queue itself might be deinitialized by the time execution reaches the `configAssert( pxQueue );` in `xQueueGenericSend`, although this error path is currently masked by #217.

fhars:
The same thing in `OTA_Suspend`:
```
    /* Check if OTA Agent is running. */
    if( otaAgent.state != OtaAgentStateStopped )
    {
        /* Stop the request timer. */
        ( void ) otaAgent.pOtaInterface->os.timer.stopTimer( OtaRequestTimer );
```
and in OTA_ActivateNewImage (this time with the insufficient check):
```
    if( ( otaAgent.pOtaInterface != NULL ) && ( otaAgent.pOtaInterface->pal.activate != NULL ) )
```

divekarshubham:
Thank you for bringing these issues to our attention. We will work on adding further checks for these functions.

lightblu:
+1

Want to add that I believe this doesn't even need such a hard/tight preemptive race like theorized above, but could even happen with e.g. cooperative FreeRTOS timer race, because a timer may just expire when `OTA_Shutdown` is called, and the timer callback then stills gets called afterwards.

That was my theory for some time, until I realized [OTA_Shutdown does not clear started timers at all](https://github.com/aws/ota-for-aws-iot-embedded-sdk/issues/412).

_I feel bad for complaining on an open source project like this, but as this is the advertised library by AWS I have to leave the snarky comment that I'm quite disappointed that this very ugly system crashing issue has not been touched for over half a year now, while at the same time endless CBMC proofs are added for other places ( reading [CBMC verifies memory safety (which includes array bounds checks and checks for the safe use of pointers)](https://www.cprover.org/cbmc/) after having to figure out that one feels funny..)_

archigup:
[...]

PR #440, opened 2022-06-10
Keep event/callback functions pointer in OTA_Shutdown to avoid accessing NULL pointer.

### LINKED ISSUE #219: Race condition and undefined behaviour in `OTA_Shutdown`

ActoryOu:
Keep event/callback functions pointer in OTA_Shutdown to avoid accessing NULL pointer.

Description
-----------

For #218 , we need a method to make sure there is no NULL function pointer.
1. Check if OTA initialized in public API OTA_SignalEvent and OTA_EventProcessingTask.
2. Do not erase the functions pointer, pOtaInterface and OtaAppCallback, in OTA_Shutdown.
3. Only create event queue in the first OTA_Init() --> so we don't need to deinit it for #217 
4. Only change the OTA state in task running OTA_EventProcessingTask. --> fix #219 

Checklist:
----------

### DIFF
--- a/source/ota.c
+++ b/source/ota.c
@@ -2831,20 +2831,6 @@ static void agentShutdownCleanup( void )
      * Clear active job name.
      */
     ( void ) memset( otaAgent.pActiveJobName, 0, OTA_JOB_ID_MAX_SIZE );
-
-    /* Clear the entire agent context except for pOtaInterface and OtaAppCallback
-     * to avoid accessing NULL function pointer if other task is running. Don't
-     * reset isOtaInterfaceInited to ensure os.event won't be created again in the
-     * next OTA_Init. */
-    ( void ) memset( &otaAgent.pThingName, 0, sizeof( otaAgent.pThingName ) );
-    ( void ) memset( &otaAgent.fileContext, 0, sizeof( otaAgent.fileContext ) );
-    otaAgent.fileIndex = 0;
-    otaAgent.serverFileID = 0;
-    otaAgent.imageState = OtaImageStateUnknown;
-    otaAgent.numOfBlocksToReceive = 0;
-    ( void ) memset( &otaAgent.statistics, 0, sizeof( otaAgent.statistics ) );
-    otaAgent.requestMomentum = 0;
-    otaAgent.unsubscribeOnShutdown = 0;
 }
 
 /*
--- a/source/ota.c
+++ b/source/ota.c
@@ -3080,6 +3080,7 @@ static void initializeAppBuffers( OtaAppBuffer_t * pOtaBuffer )
     }
     else
     {
+        otaAgent.fileContext.pFilePath = NULL;
         otaAgent.fileContext.filePathMaxSize = 0;
     }
 
@@ -3091,6 +3092,7 @@ static void initializeAppBuffers( OtaAppBuffer_t * pOtaBuffer )
     }
     else
     {
+        otaAgent.fileContext.pCertFilepath = NULL;
         otaAgent.fileContext.certFilePathMaxSize = 0;
     }
 
@@ -3102,6 +3104,7 @@ static void initializeAppBuffers( OtaAppBuffer_t * pOtaBuffer )
     }
     else
     {
+        otaAgent.fileContext.pStreamName = NULL;
         otaAgent.fileContext.streamNameMaxSize = 0;
     }
 
@@ -3113,6 +3116,7 @@ static void initializeAppBuffers( OtaAppBuffer_t * pOtaBuffer )
     }
     else
     {
+        otaAgent.fileContext.pDecodeMem = NULL;
         otaAgent.fileContext.decodeMemMaxSize = 0;
     }
 
@@ -3124,6 +3128,7 @@ static void initializeAppBuffers( OtaAppBuffer_t * pOtaBuffer )
     }
     else
     {
+        otaAgent.fileContext.pRxBlockBitmap = NULL;
         otaAgent.fileContext.blockBitmapMaxSize = 0;
     }
 
@@ -3135,6 +3140,7 @@ static void initializeAppBuffers( OtaAppBuffer_t * pOtaBuffer )
     }
     else
     {
+        otaAgent.fileContext.pUpdateUrlPath = NULL;
         otaAgent.fileContext.updateUrlMaxSize = 0;
     }
 
@@ -3146,6 +3152,7 @@ static void initializeAppBuffers( OtaAppBuffer_t * pOtaBuffer )
     }
     else
     {
+        otaAgent.fileContext.pAuthScheme = NULL;
         otaAgent.fileContext.authSchemeMaxSize = 0;
     }
 }
@@ -3288,13 +3295,7 @@ OtaState_t OTA_Shutdown( uint32_t ticksToWait,
                 "ticks=%u",
                 ticks ) );
 
-    if( otaAgent.state == OtaAgentStateInit )
-    {
-        /* When in init state, the OTA state machine is not running yet. So directly set state to
-         * stopped. */
-        otaAgent.state = OtaAgentStateStopped;
-    }
-    else if( ( otaAgent.state != OtaAgentStateStopped ) && ( otaAgent.state != OtaAgentStateShuttingDown ) ) /* LCOV_EXCL_BR_LINE */
+    if( ( otaAgent.state != OtaAgentStateStopped ) && ( otaAgent.state != OtaAgentStateShuttingDown ) ) /* LCOV_EXCL_BR_LINE */
     {
         otaAgent.unsubscribeOnShutdown = unsubscribeFlag;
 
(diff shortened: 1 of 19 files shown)


source/ota.c
```diff
@@ -1336,8 +1337,19 @@ static OtaErr_t shutdownHandler( const OtaEventData_t * pEventData )
     /* If we're here, we're shutting down the OTA agent. Free up all resources and quit. */
     agentShutdownCleanup();
 
-    /* Clear the entire agent context. This includes the OTA agent state. */
-    ( void ) memset( &otaAgent, 0, sizeof( otaAgent ) );
+    /* Clear the entire agent context except for pOtaInterface and OtaAppCallback
+     * to avoid accessing NULL function pointer if other task is running. Don't
+     * reset isOtaInterfaceInited to ensure os.event won't be created again in the
+     * next OTA_Init. */
+    ( void ) memset( &otaAgent.pThingName, 0, sizeof( otaAgent.pThingName ) );
```
ravibhagavandas, line 1344: Initializing the pointers and other fields to zero value would be better in terms of readability here.
```
otaAgent.pThingName = NULL;
otaAgent.fileContext = NULL;
....
otaAgent.statistics = 0;
```

source/ota.c
```diff
@@ -2968,14 +2989,21 @@ void OTA_EventProcessingTask( void * pUnused )
 {
     ( void ) pUnused;
 
-    /*
-     * OTA Agent is ready to receive and process events so update the state to ready.
-     */
-    otaAgent.state = OtaAgentStateReady;
-
-    while( otaAgent.state != OtaAgentStateStopped )
+    if( otaAgent.state == OtaAgentStateStopped )
```
ravibhagavandas, line 2992: Does this change the behavior of the existing OTA state machine? As per the current code, we can restart the OTA task after its shutdown?

source/ota.c
```diff
@@ -2984,35 +3012,44 @@ bool OTA_SignalEvent( const OtaEventMsg_t * const pEventMsg )
...
     OtaOsStatus_t err = OtaOsSuccess;
 
-    /* Check if file block received and update statistics.*/
-    if( pEventMsg->eventId == OtaAgentEventReceivedFileBlock )
+    /* Check if OTA library is initialized. */
+    if( otaAgent.state == OtaAgentStateStopped )
     {
-        otaAgent.statistics.otaPacketsReceived++;
+        retVal = false;
+        LogDebug( ( "Send event failed, OTA is stopped." ) );
```
ravibhagavandas, line 3019: This should be an error log.

source/ota.c
```diff
@@ -3170,7 +3215,19 @@ OtaErr_t OTA_Init( OtaAppBuffer_t * pOtaBuffer,
...
+        {
+            ( void ) otaAgent.pOtaInterface->os.event.init( NULL );
+            otaAgent.isOtaInterfaceInited = 1;
+        }
+        else
+        {
+            /* Drop all events that created after OTA_Shutdown. */
+            while( otaAgent.pOtaInterface->os.event.recv( NULL, &eventMsg, 0 ) == OtaOsSuccess )
+            {
+                LogDebug( ( "Event(%d) is dropped.\r\n", eventMsg.eventId ) );
```
ravibhagavandas, line 3228: This can be a warning log to indicate the event ID dropped.

source/ota.c
```diff
@@ -3339,7 +3382,7 @@ OtaErr_t OTA_ActivateNewImage( void )
      * and not return unless there is a problem within the PAL layer. If it does return,
      * output an error message. The device may need to be reset manually.
      */
-    if( ( otaAgent.pOtaInterface != NULL ) && ( otaAgent.pOtaInterface->pal.activate != NULL ) )
+    if( ( otaAgent.state != OtaAgentStateStopped ) && ( otaAgent.pOtaInterface != NULL ) && ( otaAgent.pOtaInterface->pal.activate != NULL ) )
```
ravibhagavandas, line 3385: I think having the check for `OtaAgentStateStopped` does not add much value as the shutdown happens concurrently in OTA agent thread.

source/portable/os/ota_os_freertos.c
```diff
@@ -145,10 +144,13 @@ OtaOsStatus_t OtaReceiveEvent_FreeRTOS( OtaEventContext_t * pEventCtx,
     {
         otaOsStatus = OtaOsEventQueueReceiveFailed;
 
-        LogError( ( "Failed to receive event from OTA Event Queue: "
-                    "xQueueReceive returned error: "
-                    "OtaOsStatus_t=%i ",
-                    otaOsStatus ) );
+        if( retVal != errQUEUE_EMPTY )
```
ravibhagavandas, line 147: xQueueReceive either returns `pdTRUE` if there is an item is retrieved from queue or `pdFALSE` if there is no item for a timeout period. So I think checking for `errQUEUE_EMPTY` is not useful for this API.

source/ota.c
```diff
@@ -3170,7 +3215,19 @@ OtaErr_t OTA_Init( OtaAppBuffer_t * pOtaBuffer,
...
-        ( void ) otaAgent.pOtaInterface->os.event.init( NULL );
+        if( !otaAgent.isOtaInterfaceInited )
+        {
+            ( void ) otaAgent.pOtaInterface->os.event.init( NULL );
+            otaAgent.isOtaInterfaceInited = 1;
+        }
+        else
+        {
+            /* Drop all events that created after OTA_Shutdown. */
+            while( otaAgent.pOtaInterface->os.event.recv( NULL, &eventMsg, 0 ) == OtaOsSuccess )
```
ravibhagavandas, line 3226: Could we have a new function  like xQueueReset() which can reset the queue atomically ?

source/ota.c
ActoryOu: No, this doesn't change the behavior of the existing OTA state. The OTA can be re-started after shutdown by calling OTA_Init() -> OTA_EventProcessingTask().

source/ota.c
ActoryOu: Fixed in the latest commit.

source/ota.c
ActoryOu: Fixed in latest commit.

source/ota.c
ActoryOu: Fixed in latest commit.

source/ota.c
ActoryOu: Function resetEventQueue added.

source/portable/os/ota_os_freertos.c
ActoryOu: Removed the condition and add a config for polling timeout to avoid printing lots of error log. Set the polling timeout to 1s by default.

source/portable/os/ota_os_freertos.c
```diff
@@ -145,7 +144,7 @@ OtaOsStatus_t OtaReceiveEvent_FreeRTOS( OtaEventContext_t * pEventCtx,
     {
         otaOsStatus = OtaOsEventQueueReceiveFailed;
 
-        LogError( ( "Failed to receive event from OTA Event Queue: "
+        LogError( ( "Failed to receive event or timeout from OTA Event Queue: "
```
ravibhagavandas, line 147: Now that queue receive has a timeout value can this cause noisy logs when the OTA agent is idle?

source/ota.c
```diff
@@ -3112,6 +3161,16 @@ static void initializeLocalBuffers( void )
     otaAgent.fileContext.pSignature = &sigBuffer;
 }
 
+static void resetEventQueue( void )
+{
+    OtaEventMsg_t eventMsg = { 0 };
+
+    while( otaAgent.pOtaInterface->os.event.recv( NULL, &eventMsg, 0 ) == OtaOsSuccess )
```
ravibhagavandas, line 3168: This could cause a memory leak if the events queued are those of job or data events for which user has created a buffer. We should probably handle the events with a user callback so that application can free those buffers.

source/portable/os/ota_os_freertos.c
ActoryOu: Set it to LogDebug to reduce noisy logs.

source/ota.c
ActoryOu: Call handleUnexpectedEvents to handle that scenarios. And checked that no other events need to be handled.

paulbartell (approved):

PR #441, opened 2022-06-10
Feature: Add user callable agent cycler

dachalco:
Description
-----------
Some users may not want to spawn a new thread just to run ota agent. To accommodate, a single-cycle function has been added that effectively runs a single iteration of the loop in `OTA_EventProcessingTask`.

Currently `OtaAgentStateReady` is used by the library to know when the ota processor task has started. This notion of task readiness is retained with the new function, where the agent is not `OtaAgentStateReady` until the `OTA_EventProcess` has been called for the first time in it's new, user defined, daemon.

State is returned from `OTA_EventProcess` as a potential for users to react to, or perhaps log, the ota agent state upon `OTA_EventProcess` return. 

Caveats
-----------
Unfortunately, the library does not provide thread sync mechanisms in its PAL OSAL, so using both `OTA_EventProcess` and `OTA_EventProcessingTask`  in an application is _not thread safe_. Perhaps an OTA user config should be added to make the two mutually exclusive in builds?

Checklist:
----------

- [x] ~~I have read and applied the rules stated in CONTRIBUTING.md.~~ n/a No such file.

### DIFF
--- a/source/ota.c
+++ b/source/ota.c
@@ -511,8 +511,7 @@ static OtaAgentContext_t otaAgent =
     0,                    /* requestMomentum */
     NULL,                 /* pOtaInterface */
     NULL,                 /* OtaAppCallback */
-    1,                    /* unsubscribe flag */
-    false                 /* agentStarted flag */
+    1                     /* unsubscribe flag */
 };
 
 /**
@@ -2973,7 +2972,6 @@ void OTA_EventProcessingTask( void * pUnused )
      * OTA Agent is ready to receive and process events so update the state to ready.
      */
     otaAgent.state = OtaAgentStateReady;
-    otaAgent.agentStarted = true;
 
     while( otaAgent.state != OtaAgentStateStopped )
     {
@@ -2981,16 +2979,15 @@ void OTA_EventProcessingTask( void * pUnused )
     }
 }
 
-OtaState_t OTA_EventProcess( void * pUnused )
+OtaState_t OTA_EventProcess( void )
 {
-    ( void ) pUnused;
-
-    /* Since user can single cycle the agent, we cannot repeatedly change state back to OtaAgentStateReady
-     * as it may interfere with an ongoing OTA. However, OtaAgentStateReady must still indicate agent readiness.*/
-    if( !otaAgent.agentStarted )
+    /* Technically, setting readiness on this condition does not match conditions which EventProcessingTask
+     * sets readiness, as this function only sets readiness with preceding call to OTA_Init, whereas
+     * EventProcessingTask does not, and sets readiness independent of OTA_Init.
+     */
+    if( otaAgent.state == OtaAgentStateInit )
     {
         otaAgent.state = OtaAgentStateReady;
-        otaAgent.agentStarted = true;
     }
 
     if( otaAgent.state != OtaAgentStateStopped )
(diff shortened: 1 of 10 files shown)


source/ota.c
```diff
@@ -2972,13 +2973,34 @@ void OTA_EventProcessingTask( void * pUnused )
...
     otaAgent.state = OtaAgentStateReady;
+    otaAgent.agentStarted = true;
 
     while( otaAgent.state != OtaAgentStateStopped )
     {
         receiveAndProcessOtaEvent();
     }
 }
 
+OtaState_t OTA_EventProcess( void * pUnused )
```
ravibhagavandas, line 2984: Any reason for keeping the `pUnused` parameter ?

source/ota.c
```diff
@@ -2972,13 +2973,34 @@ void OTA_EventProcessingTask( void * pUnused )
...
+{
+    ( void ) pUnused;
+
+    /* Since user can single cycle the agent, we cannot repeatedly change state back to OtaAgentStateReady
+     * as it may interfere with an ongoing OTA. However, OtaAgentStateReady must still indicate agent readiness.*/
+    if( !otaAgent.agentStarted )
```
ravibhagavandas, line 2990: Could we check for `(otaAgent.state == OtaAgentStateInit )` instead of adding new variable `agentStarted` ?
Are there other transitions to ready state, for example after stopping OTA can the agent be started again?

source/ota.c
dachalco: Just to match the existing `OTA_EventProcessingTask`, figured it was stubbed there so any future usage wouldn't result in changes to `ota.h`

source/ota.c
dachalco: I don’t think the checking for `state == OtaAgentStateInit` will work. `EventProcessingTask` sets state to `OtaAgentStateReady` irrespective of any conditions. From what I’ve seen, `OtaAgentStateReady` is strictly for indicating when ota thread has started running. To match that criteria we couldn’t check for state == `OtaAgentStateInit` since agent is initialized as `OtaAgentStateStopped` (see ota.c:498). So at first call to `OTA_EventProcess` in user OTA thread, `state != OtaAgentStateInit` unless user has called `OTA_Init` _before_ user ota daemon has had a chance to call `OTA_EventProcess`.

ravibhagavandas: IMO `OTA_EventProcess` should never be called before `OTA_Init()` as it sets the PAL functions and thing name etc.. in OTA Init.

dachalco: I agree, but it doesn't discount that you _can_ technically create the `OTA_EventProcessingTask` before calling `OTA_Init` and app would still work. Matching readiness based on this condition != conditions by which `OTA_EventProcessingTask` sets state to `OtaAgentStateReady`. 

Though the mentioned is an _incorrect_ flow, at least in that it exclusively produces error logs, it does _work_ since OTAs would work nominally after `OTA_Init` _is_ called.

dachalco: Regardless of that technical tidbit, I do like like the suggested change. Added it in the latest commit, with a comment stating that the readiness it indicates is only _approximately_ equal to readiness indicated by `OTA_EventProcessingTask`.

ravibhagavandas (approved):

cobusve (approved):

Issue #220, opened 2021-05-25
AppCallback API usage in all examples is not const correct


fhars:
The second parameter of `OtaAppCallback_t` is defined as `const void * pData` , but then the only documented uses of that parameter are to use it for memory management purposes and cast the constness away:

https://github.com/FreeRTOS/FreeRTOS/blob/c134a581153c1b137af2af2e93564c495977872f/FreeRTOS-Plus/Demo/AWS/Ota_Windows_Simulator/Ota_Over_Mqtt_Demo/DemoTasks/OtaOverMqttDemoExample.c#L836

https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/5eec3f03f420b904e1c041e5bda3eeebb6f5744b/demos/ota/ota_demo_core_mqtt/ota_demo_core_mqtt.c#L664

Also, if some messages happen to be in the queue the moment the agent decides to go into shutdown, the callback is never called for these packages and the buffers leak.

pvyawaha:
Hello , 

We are working on adding more documentation and examples for the `OtaAppCallback_t `. The buffers are in the demos and not the library so it can be cleared in the demo/application and we will add that in the ota demos in the repository.

ActoryOu:
Hi,
Thanks again for contribution.
We removed the const property in the OtaAppCallback_t function prototype with PR #443.
And we also provided more description for this function usage.

I'm going to close this issue. Any feedback is welcome.

Thanks.

PR #443, opened 2022-06-13
Add more description on OtaAppCallback_t and remove const property on pData.

ActoryOu:
Add more description on OtaAppCallback_t to help user get familiar with it sooner, and remove const property on pData to avoid misunderstanding.

Description
-----------

- Add more description on OtaAppCallback_t to help user get familiar with it faster.
- Remove const property on pData to avoid misunderstanding.
  - As #220 mentioned, pData is used for memory management in demo, but we declared it as const may make users misunderstand it. So, remove the property.

Checklist:
----------

### DIFF
--- a/source/include/ota.h
+++ b/source/include/ota.h
@@ -226,9 +226,16 @@ typedef struct OtaJobDocument
  *
  * The callback function is called with one of the following arguments:
  *
- *      OtaJobEventActivate      OTA update is authenticated and ready to activate.
- *      OtaJobEventFail          OTA update failed. Unable to use this update.
- *      OtaJobEventStartTest     OTA job is now ready for optional user self tests.
+ *      OtaJobEventActivate           OTA update is authenticated and ready to activate.
+ *      OtaJobEventFail               OTA update failed. Unable to use this update.
+ *      OtaJobEventStartTest          OTA job is now ready for optional user self tests.
+ *      OtaJobEventProcessed          Event OTA got from user has been handled.
+ *      OtaJobEventSelfTestFailed     OTA update failed in self-test.
+ *      OtaJobEventParseCustomJob     OTA got an unknown job from cloud, need user to check
+ *                                    if it's a custom job.
+ *      OtaJobEventReceivedJob        OTA event when a new valid job is received.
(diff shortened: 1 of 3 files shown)


source/include/ota.h
```diff
@@ -238,12 +245,47 @@ typedef struct OtaJobDocument
...
+ * OtaJobEventSelfTestFailed|NULL|nothing
+ * OtaJobEventParseCustomJob|OtaJobDocument_t|pJobId, jobIdLength, pJobDocJson, and jobDocLength
+ * OtaJobEventReceivedJob|OtaJobDocument_t|pJobId, jobIdLength, pJobDocJson, jobDocLength, and fileTypeId
+ * OtaJobEventUpdateComplete|OtaJobDocument_t|status, reason and subReason
+ * OtaJobEventNoActiveJob|NULL|nothing
+ *
  */
 typedef void (* OtaAppCallback_t)( OtaJobEvent_t eEvent,
-                                   const void * pData );
+                                   void * pData );
```
ravibhagavandas, line 288: Would this cause an incompatibility with the applications using ota library as in it now generates a warning of type incompatibility while assignment.

Should we also update all the demo applications with the change?

source/ota.c
```diff
@@ -1008,7 +1008,7 @@ static OtaErr_t processJobHandler( const OtaEventData_t * pEventData )
     }
 
     /* Application callback for event processed. */
-    callOtaCallback( OtaJobEventProcessed, ( const void * ) pEventData );
+    callOtaCallback( OtaJobEventProcessed, ( void * ) pEventData );
```
ravibhagavandas, line 1011: No need of explicit type case to ( void * ). Same at other places in code.

source/include/ota.h
ActoryOu: Yes, this change would cause compile warning for the data structure. And we need to update all the demo applications with the change when we merge it.

source/ota.c
ActoryOu: I need to keep this cast for pEventData because it's const as input, or there is a build warning.
I removed some because they're not necessary.

ravibhagavandas (approved):

paulbartell (approved):

Issue #214, opened 2021-05-04
Issues with integration into an application


boraozgen:
I am trying to integrate the OTA agent to an application. While the demo works fairly well on its own, I believe that the integration into an application is not yet ideal. The missing features that occured to me:

1. Not possible to defer the download after receiving an update job to a future time, e.g. using a `StartDownload` command.
2. Not possible to get the result of the job document after a job document is received. Does an update job exist or is the job document empty? Can we move on with the application or is the download in progress? This is critical for applications which would like to suspend other operations while the download is in progress.
3. How to figure out when an external `MQTT_ProcessLoop` is required to continue with the download? For systems with long socket receive timeouts, calling ProcessLoop (i.e. socket recv) repeatedly causes unnecessary delays and would be nice to avoid if the OTA agent does not expect a response at the moment.

pvyawaha:
Hello,

> 1.Not possible to defer the download after receiving an update job to a future time, e.g. using a StartDownload command.

When a  job document is received  the application callback ( otaAppCallback in demo ) has an event `OtaJobEventReceivedJob `. This event has data associated with it of type `OtaJobDocument_t` which has fields - 
`pJobDocJson ` Pointer to entire job document ( JSON )
`jobDocLength` Size of the job document 
`pJobId ` pointer to Job ID
`jobIdLength` Job ID length
`fileTypeId ` File type ID

The callback event is called when the received job document json is valid and has execution key. When this even is received application can wait for an event if it is short wait like winding down some application specific task, or we can suspend and resume OTA Agent when application is ready or even shutdown and restart using the OTA Agent APIs.

>2. Not possible to get the result of the job document after a job document is received. Does an update job exist or is the job document empty? Can we move on with the application or is the download in progress? This is critical for applications which would like to suspend other operations while the download is in progress.  

The `OTA_GetState ` gives the state of type `OtaState_t` which has four states for file and download operations - 

    `OtaAgentStateCreatingFile` - File is being created to receive the update file or on embedded platforms without file system it can be erasing a flash partition.
    `OtaAgentStateRequestingFileBlock` - The OTA Agent is sending requests for file blocks to the service.
    `OtaAgentStateWaitingForFileBlock ` - The OTA Agent is waiting for response from the service for file blocks.
    `OtaAgentStateClosingFile` - File is being closed and crypto signature validated as the download is complete.

[...]

PR #445, opened 2022-06-14
Notify user that an OTA download is going to start.

ActoryOu:
Notify user that an OTA download is going to start.

Description
-----------

Callback to user when a new OTA job is received.

Checklist:
----------

### DIFF
--- a/docs/doxygen/include/size_table.md
+++ b/docs/doxygen/include/size_table.md
@@ -9,8 +9,8 @@
     </tr>
     <tr>
         <td>ota.c</td>
-        <td><center>8.3K</center></td>
-        <td><center>7.5K</center></td>
+        <td><center>8.4K</center></td>
+        <td><center>7.6K</center></td>
     </tr>
     <tr>
         <td>ota_interface.c</td>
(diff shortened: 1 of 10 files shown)


ravibhagavandas (commented):
Any specific reason for adding this new notification type ?
Can we pass any useful information along with the event type ?
Can we add unit tests for the change?

ActoryOu:
In #214 description, user needs to know what's the job handling in OTA library.
Currently OTA callback to user when self-test started (OtaJobEventStartTest), custom job (OtaJobEventParseCustomJob), or invalid job (OtaJobParseErrNoActiveJobs), but there is no notification for normal OTA job started.

> For the startup after the connection, I signal an OtaAgentEventStart event. This requests a job document (makes sense for self-test), but in case a new update exists, it starts downloading the update immediately. To stop the update, the agent must then be suspended explicitly, and the point to do this is unclear. In case of a self-test, we would like to receive the document and keep working (multiple ProcessLoops are required to finish self-test) until no new messages are received and we are still in OtaAgentStateWaitingForJob state. Therefore I cannot suspend the agent immediately after starting it. I had to build a relatively complicated logic to handle this. I think adding additional states like "update available" and "performing self-test" would be better.

I think we can provide a job doc for user to know the job ID, I'll add it. And I'll try to add unit tests.

Issue #460, opened 2022-11-07
[BUG] Issue with longer thing names


LozGMK:
Apart from the thing name length issue described in this other issue: [[Feature Request] Misleading default config for otaconfigMAX_THINGNAME_LEN](https://github.com/aws/ota-for-aws-iot-embedded-sdk/issues/456), there is another limit to thing name length in this code.

In `ota-for-aws-iot-embedded-sdk/source/ota_mqtt.c`, the thing name is added into string `pPayloadParts[ 3 ]`, which is part of a message published to AWS topic `$aws/things/[thing_name]/jobs/notify-next` that ends up something like `{"clientToken":"[decimal_value_of_reqCounter_in_a_string]:[thing_name]"}`. Since the decimal value of `reqCounter` can be up to 10 characters, the  thing name can only be 64 - 10 -1 = 53 chars before there is (potentially) a problem with this request, as AWS complains (and does not send back the requested data) if a client token longer than 64 characters is used.

If the intention is to place a unique value in `clientToken` (to tie the request and response together), then surely the value of `reqCounter` is sufficient, without also including the thing name?

As a side note, the value of `reqCounter` is turned into a decimal string using function `stringBuilderUInt32Decimal()`, which returns an empty string when it is sent a value of `0`, rather than `"0"`, as might be expected.

That function may cause issues for other uses of the function, such as with `versionMajorString`, `versionMinorString` and `versionBuildString`.

Tested on an ESP32.

AniruddhaKanhere:
Hello @LozGMK,

Thank you for bringing this to our attention. I'll inform the development team about this and will get back to you soon.

Thank you for your patience,
Aniruddha

AniruddhaKanhere:
@dachalco and @LozGMK,

I have created a PR to fix the string builder issue: https://github.com/aws/ota-for-aws-iot-embedded-sdk/pull/461 can you please take a look?

LozGMK:
I've had a look, put the same fix into my own code, and it seems to work fine.

LozGMK:
Were there any thoughts on the other issues?

Just to clarify, those are:
- that there is a misleading (and excessively limiting) default config value for `otaconfigMAX_THINGNAME_LEN`, limiting thing name length (value should be `128` - I think this is specified in two places in the code, though one may be `#define`d out)
[...]

PR #461, opened 2022-11-10
Fix stringBuilderUInt32Decimal function when 0 is provided as the input

AniruddhaKanhere:
Description
-----------
As outlined in this issue: https://github.com/aws/ota-for-aws-iot-embedded-sdk/issues/460, the `stringBuilderUInt32Decimal` function returns an empty string when the value passed to it is 0 instead of returning "0\0" (0 in ASCII followed by the NULL character).

Checklist:
----------

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
         uses: FreeRTOS/CI-CD-Github-Actions/complexity@main
         with:
           path: ./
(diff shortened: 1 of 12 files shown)


LozGMK (commented):
That seems to work fine to me.

kstribrnAmzn (approved):

paulbartell (approved):

PR #472, opened 2023-01-06
fix: warnings with cross-GCC in ota_cbor.c

cookpate:
# fix: warnings with cross GCC in ota_cbor.c

Description
-----------

This commit adds a helper function in ota_cbor.c to call `cbor_value_get_int64()` and perform the narrowing conversion from `int64_t` directly to `int32_t` ourselves.

TinyCBOR takes `int*` in `cbor_value_get_int()`. We currently pass `int32_t*`, which is implicitly converted to `int*` on platforms where `int` is not a 32-bit signed integer. This results in undefined behavior. Aside from UB, this code may cast to a narrower type before again casting to a wider type, which loses precision.

Checklist:
----------


cookpate:
It looks like for the CBMC proof I'll need to add back in the narrow conversion check & unit test.

cookpate:
This warning is resolved by (#485)

PR #475, opened 2023-02-17
Fix stringBuilderUInt32Decimal to handle 0 value

kstribrnAmzn:
Description
-----------
Previously the stringBuilder method would return
and empty string when the integer value of '0' was supplied. With this change, the string representation "0" is returned.

Checklist:
----------

- [ x ] I have tested my changes. No regression in existing tests.
- [ x ] My code is formatted using Uncrustify.
- [ x ] I have read and applied the rules stated in CONTRIBUTING.md. 

### DIFF
--- a/source/ota_mqtt.c
+++ b/source/ota_mqtt.c
@@ -297,7 +297,7 @@ static size_t stringBuilderUInt32Decimal( char * pBuffer,
 
     /* Assert if there is not enough buffer space. */
 
-    assert( bufferSizeBytes >= U32_MAX_LEN );
+    assert( bufferSizeBytes >= U32_MAX_LEN + 1 );
     ( void ) bufferSizeBytes;
 
     do
@@ -315,7 +315,6 @@ static size_t stringBuilderUInt32Decimal( char * pBuffer,
     }
 
     pDest[ size ] = '\0';
-    size++;
     return size;
 }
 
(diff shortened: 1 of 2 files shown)


paulbartell (dismissed):
Looks reasonable to me. I would be more comfortable if a test were included though.

kstribrnAmzn:
For stringBuilder methods there were two bugs...
1. The buffer was actually undersized. The buffer needs to be one larger than UINT_32_MAX_SIZE (which is 10 characters) to account for the null terminating character. This null terminating character is required in the later `stringBuilder` function since it uses `strncat` which requires null terminated strings ([see doc](https://www.ibm.com/docs/en/i/7.1?topic=functions-strncat-concatenate-strings))
2. The returned `size` WAS accounting for the null terminating character which goes against the function doxygen documentation.

paulbartell (dismissed):

source/ota_mqtt.c
```diff
@@ -282,6 +282,8 @@ static size_t stringBuilder( char * pBuffer,
         curLen += thisLength;
```
AniruddhaKanhere, line 282: Beyond the scope of this PR, but I dislike that we are only asserting (on line 279) on the string length not exceeding the buffer size. In case assert is not defined, we'll simply cause memory corruption.

AniruddhaKanhere (approved):
Thank you @kstribrnAmzn for this PR.

I am Approving this with one comment regarding asserts and not having a fail-safe in case of absence of asserts.

source/ota_mqtt.c
kstribrnAmzn: Is there a standard way we avoid this? I'm happy to incorporate that change.

AniruddhaKanhere (approved):

PR #477, opened 2023-03-08
OTA Bitmap Field Size Comparison

### LINKED ISSUE #476: [BUG] Memory corruption if OTA_MAX_BLOCK_BITMAP_SIZE is too small for the file
**Describe the bug**
As per issue on re:Post: https://repost.aws/questions/QU4tQPeyESRUqKmMXC__TcYw/large-mqtt-ota-file-transfer-fails-at-block-1024

The customer appeared to suffer a memory corruption when `OTA_MAX_BLOCK_BITMAP_SIZE` was too small for their file size. Problem was resolved when `OTA_MAX_BLOCK_BITMAP_SIZE` was set to sufficient size.

**Host**
ESP32 with FreeRTOS.

**To Reproduce**
Do an OTA with `OTA_MAX_BLOCK_BITMAP_SIZE` set too small for your file size.

**Expected behavior**
OTA library should check the file size before beginning the file transfer, and fail gracefully if the file is too big. No memory corruption should occur.

**Additional context**
I haven't personally taken the time to reproduce the problem. From perusing the code, I couldn't see the library protecting itself in this situation. Discussed the problem with Soren before raising this issue.

Skptak:
Add a check into the parseJobDoc() function to see if the OTA transfer will require more blocks than the bitmap has declared.

This pull request was created to fix #476 

### DIFF
--- a/source/ota.c
+++ b/source/ota.c
@@ -2367,6 +2367,11 @@ static OtaFileContext_t * parseJobDoc( const JsonDocParam_t * pJsonExpectedParam
     }
     else
     {
+        if( pFileContext->blocksRemaining > OTA_MAX_BLOCK_BITMAP_SIZE )
+        {
+            err = OtaJobParseErrBadModelInitParams;
+        }
+
         parseError = parseJSONbyModel( pJson, messageLength, &otaJobDocModel );
 
         if( parseError == DocParseErrNone )
--- a/source/ota.c
+++ b/source/ota.c
@@ -2365,13 +2365,12 @@ static OtaFileContext_t * parseJobDoc( const JsonDocParam_t * pJsonExpectedParam
     {
         err = OtaJobParseErrBadModelInitParams;
     }
+    else if( pFileContext->blocksRemaining > OTA_MAX_BLOCK_BITMAP_SIZE )
+    {
+        err = OtaJobParseErrBadModelInitParams;
+    }
     else
     {
-        if( pFileContext->blocksRemaining > OTA_MAX_BLOCK_BITMAP_SIZE )
-        {
-            err = OtaJobParseErrBadModelInitParams;
-        }
-
         parseError = parseJSONbyModel( pJson, messageLength, &otaJobDocModel );
 
         if( parseError == DocParseErrNone )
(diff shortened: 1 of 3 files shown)


source/ota.c
```diff
@@ -2365,6 +2365,10 @@ static OtaFileContext_t * parseJobDoc( const JsonDocParam_t * pJsonExpectedParam
     {
         err = OtaJobParseErrBadModelInitParams;
     }
+    else if( pFileContext->blocksRemaining > OTA_MAX_BLOCK_BITMAP_SIZE )
+    {
+        err = OtaJobParseErrBadModelInitParams;
```
kstribrnAmzn, line 2370:
```suggestion
        LogWarn( ( "OTA size (%lu blocks) greater than can be tracked. Increase `OTA_MAX_BLOCK_BITMAP_SIZE`", pFileContext->blocksRemaining ) );
        err = OtaJobParseErrBadModelInitParams;
```

Let's add a warning message here so a resolution is clear to the user.

test/unit-test/ota_utest.c
```diff
@@ -3528,6 +3528,25 @@ void test_OTA_parseJobFailsNullJsonDocument()
...
+        { OTA_JSON_JOB_ID_KEY, OTA_JOB_PARAM_REQUIRED, U16_OFFSET( OtaFileContext_t, pJobName ), U16_OFFSET( OtaFileContext_t, jobNameMaxSize ), UINT16_MAX },
+    };
+
+    /* The document structure has an invalid value for ModelParamType_t. */
+    otaAgent.fileContext.blocksRemaining = OTA_MAX_BLOCK_BITMAP_SIZE + 1;
+    otaInitDefault();
+    pContext = parseJobDoc( otaCustomJobDocModelParamStructure, 1, JOB_DOC_A, strlen( JOB_DOC_A ), &updateJob );
+
+    TEST_ASSERT_NULL( pContext );
+    TEST_ASSERT_EQUAL( false, updateJob );
```
kstribrnAmzn, line 3546: Nit pick - I like it when the variable isn't initialized to what you expect later on that way you know the function you called updated the variable.

source/ota.c
Skptak: Good call! I actually had wanted to do this but since the if statement before it didn't I thought it might be odd that we only add the error here. Now realizing that the `initDocModel()` function has its own logging.

test/unit-test/ota_utest.c
Skptak: So you want me to remove the `otaAgent.fileContext.blocksRemaining = 0;` line? 
I don't have a problem with that and can agree with your thought process. The  reason I thought to do it was since it's a global static was worried another test might then run with the incorrect value.

Issue #491, opened 2023-06-03
Read outside of Json buffer


gonzzor:
The printf format to LogDebug expects pJson to be null-terminated, but its length is given by messageLength. Thus it will read more than expected.

https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/406415ef38d9f0fbabd69735623008fd090043b3/source/ota.c#L1821

Something like this should fix the issue.
`    LogDebug(("JSON received: %.*s", messageLength, pJson));`

gmtt:
Thank you for bringing this to our attention! We will be looking further into this issue as soon as we can, thank you for your patience

ActoryOu:
Hi @gonzzor,
Sorry for late response. The PR is created to apply this change back to OTA library.

kstribrnAmzn:
The PR - https://github.com/aws/ota-for-aws-iot-embedded-sdk/pull/493

ActoryOu:
I'm closing this issue since PR's merged

PR #493, opened 2023-07-27
Print string with length check

ActoryOu:
Print string with length check

Description
-----------

As #491 mentioned, that reads might more than expected.
Submit the change to read the string with limited length.

Checklist:
----------

### DIFF
--- a/source/ota.c
+++ b/source/ota.c
@@ -1817,8 +1817,20 @@ static DocParseErr_t parseJSONbyModel( const char * pJson,
     uint16_t paramIndex = 0;
     const char * pFileParams = NULL;
     uint32_t fileParamsLength = 0;
+    static char cJsonBuffer[ OTA_MAX_JSON_STR_LEN ];
 
-    LogDebug( ( "JSON received: %.*s", messageLength, pJson ) );
+    if( messageLength < OTA_MAX_JSON_STR_LEN - 1 )
+    {
+        memcpy( cJsonBuffer, pJson, messageLength );
+        cJsonBuffer[ messageLength + 1 ] = 0;
+    }
+    else
+    {
+        memcpy( cJsonBuffer, pJson, OTA_MAX_JSON_STR_LEN - 2 );
+        cJsonBuffer[ OTA_MAX_JSON_STR_LEN - 1 ] = 0;
+    }
+
+    LogDebug( ( "JSON received: %s", cJsonBuffer ) );
 
     /* Fetch the model parameters from the DocModel*/
     pModelParam = pDocModel->pBodyDef;
--- a/source/ota.c
+++ b/source/ota.c
@@ -1818,7 +1818,7 @@ static DocParseErr_t parseJSONbyModel( const char * pJson,
     const char * pFileParams = NULL;
     uint32_t fileParamsLength = 0;
 
-    LogDebug( ( "JSON received: %s", pJson ) );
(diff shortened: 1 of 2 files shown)


kstribrnAmzn (dismissed):

moninom1 (approved):

source/ota.c
```diff
@@ -1818,7 +1818,7 @@ static DocParseErr_t parseJSONbyModel( const char * pJson,
     const char * pFileParams = NULL;
     uint32_t fileParamsLength = 0;
 
-    LogDebug( ( "JSON received: %s", pJson ) );
+    LogDebug( ( "JSON received: %.*s", messageLength, pJson ) );
```
n9wxu, line 1821: This makes an even bigger assumption that LogDebug is implemented by printf.  We may as well just use printf.

n9wxu (changes requested):
Some lightweight embedded versions of printf do not support the string length specifiers.  There needs to be a more generic solution to this problem.

source/ota.c
ActoryOu: Thanks @n9wxu,
I updated the logic here to use a 256 bytes temp buffer to store the json because the buffer is provided by application which is not controlled by us.
To avoid big local buffer in the stack, I use static memory instead.

Please help to take a look again.

Thanks.

source/ota.c
```diff
@@ -1817,8 +1817,20 @@ static DocParseErr_t parseJSONbyModel( const char * pJson,
     uint16_t paramIndex = 0;
     const char * pFileParams = NULL;
     uint32_t fileParamsLength = 0;
+    static char cJsonBuffer[ OTA_MAX_JSON_STR_LEN ];
```
kstribrnAmzn, line 1820: Why not make this `cJsonBuffer[ OTA_MAX_JSON_STR_LEN + 1]`? Then you can store the maximum message length guaranteed.

kstribrnAmzn: You could also get rid of the if/else here assuming `OTA_MAX_JSON_STR_LEN` by doing this

ActoryOu: OTA_MAX_JSON_STR_LEN(256) is not the maximum size of whole json file. So that if/else case must be there.
Here is the example output from my test. There are about 600 charactors.
```
[DEBUG] [OTA] [ota.c:1761] JSON received: {"timestamp":1690449009,"execution":{"jobId":"AFR_OTA-
tryOTA091aaa","status":"QUEUED","queuedAt":1690449008,"lastUpdatedAt":1690449008,"versionNumber":1,
"executionNumber":1,"jobDocument":{"afr_ota":{"protocols":["MQTT"],"streamname":
"AFR_OTA-b16fc240-3c43-4778-a08c-5bf691ede395","files":[{"filepath":
"/home/ousc/git/aws-iot-device-sdk-embedded-C/build/bin/0911",
"filesize":236048,"fileid":0,"certfile":"certificates/ecdsasigner.crt","sig-sha256-ecdsa":
"MEUCIQD1PBg3xq6GmknDymDa7qXDc2gg7LvU/fk5sZnCevPacwIgJB5f5F1py8HdCo7FQdUQNCQEgv1o/C7vwqIv2+QBE2s="}]}}}}
```

kstribrnAmzn (approved):

n9wxu (approved):

PR #504, opened 2023-12-14
Fix Block Bitmap Calculation

### LINKED ISSUE #503: [BUG] OTA Block size check wrong
If a network disconnect occurs and the ota task is suspended and resumed, when the job description comes back, the library checks the ` pFileContext->blocksRemaining ` against `OTA_MAX_BLOCK_BITMAP_SIZE` (128) and silently stops the ota, without  updating the job. The same issue doesn't happen if the device reboots instead of resuming the ota task. The issue also never manifests if the ota job is not suspended in the middle of job, so the number of blocks is not checked unless we get a new job document while already having one?
The issue is 2 fold, the check is wrong, it should be(ota.c:2404):
```C
    else if( pFileContext->blocksRemaining > (OTA_MAX_BLOCK_BITMAP_SIZE * BITS_PER_BYTE) )
```
Applying the change above fixes the issue. I can create a PR with the change if you can confirm that my understanding of the issue is correct.

The other is that the ota library fails silently and the job will not resume until the device reboots.

This is the log after a network reconnection:
```
I (105814) hota: Received job document: {...}
W (105074) awsota: Index: 3. OTA event id: 3
W (105080) awsota: OTA size (266 blocks) greater than can be tracked. Increase `OTA_MAX_BLOCK_BITMAP_SIZE`(128)
I (105090) awsota: Unable to initialize Job Parsing: OtaJobParseErr_t=OtaJobParseErrBadModelInitParams
I (105100) awsota: otaPal_GetPlatformImageState
E (105106) awsota: Failed to execute state transition handler: Handler returned error: OtaErr_t=OtaErrJobParserError
E (105117) awsota: Current State=[WaitingForJob], Event=[ReceivedJobDocument], New state=[CreatingFile]
```
This is all on the main branches of ota-for-aws-iot-embedded-sdk, coreMQTT, jobs etc.

BogdanTheGeek:
Fixes https://github.com/aws/ota-for-aws-iot-embedded-sdk/issues/503

Description
-----------
Multiply buffer size by element size (byte). Maybe this should use `sizeof(*pFileBitmap)`.
I am not able to run the unit tests as I only have access to a mac and the unit tests depend on `mqueue.h`.

Checklist:
----------


bradleysmith23 (dismissed):

bradleysmith23:
Thank you for updating the PR, sorry for the delay. Please take a look at the PR checks and correct the issues, you seem to be missing a closing parentheses on line 2268 in ota.c. Thank you!

BogdanTheGeek:
Sorry about that, I couldn't run the tests locally.

BogdanTheGeek:
Hi, is this going to be merged?

bradleysmith23:
Hi, @BogdanTheGeek, my apologies for the delay. Reviewing the remaining checks now.

aggarg (approved):

kar-rahul-aws (approved):

PR #510, opened 2024-04-15
CBMC proof harness type fixes

tautschnig:
Description
-----------

These changes add necessary forward declarations (or system headers includes) and address inconsistencies in how we invoke functions from CBMC proof harnesses as compared to their original signatures.

Checklist:
----------


test/cbmc/proofs/Posix_OtaReceiveEvent/Posix_OtaReceiveEvent_harness.c
```diff
@@ -35,6 +35,8 @@
 #include <poll.h>
 #include <mqueue.h>
 
+int nondet_int();
```
kstribrnAmzn, line 38: Why does this function now need to be declared?

test/cbmc/proofs/decodeBase64IndexBuffer/decodeBase64IndexBuffer_harness.c
```diff
@@ -39,11 +39,11 @@
...
 /* Declaration of the mangled name static function which is created by CBMC for static functions. */
-Base64Status_t __CPROVER_file_local_ota_base64_c_decodeBase64IndexBuffer( uint8_t * pDest,
-                                                                          const size_t destLen,
-                                                                          size_t * pResultLen,
+Base64Status_t __CPROVER_file_local_ota_base64_c_decodeBase64IndexBuffer( uint32_t * pDest,
+                                                                          uint32_t * destLen,
                                                                           const uint8_t * pEncodedData,
-                                                                          const size_t encodedLen );
+                                                                          const size_t encodedLen,
+                                                                          size_t * pResultLen );
```
kstribrnAmzn, line 46: Do the parameter names need to match the [static function](https://github.com/aws/ota-for-aws-iot-embedded-sdk/blob/2557f84b8295a30306aced925048d1073dcde968/source/ota_base64.c#L318-L322) which CBMC is overriding?

test/cbmc/proofs/freeFileContextMem/freeFileContextMem_harness.c
```diff
@@ -34,6 +34,8 @@
 extern OtaAgentContext_t otaAgent;
 extern void freeFileContextMem( OtaFileContext_t * const pFileContext );
 
+__CPROVER_bool nondet_bool();
```
kstribrnAmzn, line 37: Why does this need to be declared here? And why the special return type given the earlier example `nondet_int` was returning a standard type?

test/cbmc/proofs/getFileContextFromJob/getFileContextFromJob_harness.c
```diff
@@ -84,7 +86,7 @@ DocParseErr_t parseJobDoc( const JsonDocParam_t * pJsonExpectedParams,
 
 void getFileContextFromJob_harness()
 {
-    OtaFileContext_t * pFileContext;
+    OtaFileContext_t * pFileContext = &fileContext;
```
kstribrnAmzn, line 89: Why not get rid of this previously unused variable instead of now assigning it a value?

test/cbmc/proofs/Malloc_FreeRTOS/Malloc_FreeRTOS_harness.c
```diff
@@ -29,6 +29,8 @@
 /*  FreeRTOS includes for OTA library. */
 #include "ota_os_freertos.h"
 
+#include <stdlib.h>
```
AniruddhaKanhere, line 32: NIT: should we not include such standard libs before the headers of our own libraries as a rule of FreeRTOS?

test/cbmc/proofs/Posix_OtaReceiveEvent/Posix_OtaReceiveEvent_harness.c
tautschnig: It always had to be, it just seems no one cared about the warning.

test/cbmc/proofs/freeFileContextMem/freeFileContextMem_harness.c
tautschnig: Here, this is only used for a true/false decision. We could, however, also pick a different type (that can be converted to an `int`) in here.

test/cbmc/proofs/Malloc_FreeRTOS/Malloc_FreeRTOS_harness.c
tautschnig: Happy to adjust this if there is such a rule, I wasn't aware.

test/cbmc/proofs/decodeBase64IndexBuffer/decodeBase64IndexBuffer_harness.c
tautschnig: The names do not need to match, just the types (note that previously several parameters were missing.

kstribrnAmzn (approved):

AniruddhaKanhere (approved):

